PR #1, opened 2015-07-30
(HC-3) Port the basic ConfigValue classes

Magisus:


MikaelSmith:
We should prefix these with the ticket number now.

lib/inc/internal/abstract_config_value.hpp
```diff
@@ -9,9 +9,8 @@ namespace hocon {
     public:
         abstract_config_value(simple_config_origin origin);
 
-        virtual std::string transform_to_string();
-        virtual config_value_type value_type() = 0;
-        virtual abstract_config_value* new_copy(simple_config_origin origin) = 0;
+        virtual std::string transform_to_string() const;
+        virtual config_value_type value_type() const = 0;
```
MikaelSmith: This might not be declared, since it's in config_value. Not completely sure.

lib/inc/hocon/config_value.hpp
```diff
@@ -33,6 +33,6 @@ namespace hocon {
          *
          * @return value's type
          */
-        virtual config_value_type value_type() = 0;
+        virtual config_value_type value_type() const = 0;
```
MikaelSmith: Should this really return by value, or by reference?

lib/inc/internal/config_boolean.hpp
```diff
@@ -1,15 +1,15 @@
...
 namespace hocon {
 
     class config_boolean : public abstract_config_value {
     public:
         config_boolean(simple_config_origin origin, bool value);
-        config_value_type value_type();
-        config_boolean* new_copy(simple_config_origin origin) override;
-        std::string transform_to_string() override;
+
+        config_value_type value_type() const;
```
MikaelSmith: This seems like it should have the override keyword.

lib/inc/hocon/config_value.hpp
Magisus: How are enums best handled? I wasn't sure. config_value_type is an enum.

MikaelSmith: Oh, just an enum.

MikaelSmith:
Generally looks sensible. I need to do a closer comparison to the java classes.

lib/src/simple_config_origin.cc
```diff
@@ -1,34 +1,28 @@
...
 
 namespace hocon {
 
     simple_config_origin::simple_config_origin(string description, int line_number, int end_line_number,
-                                               origin_type org_type, string resource_or_null,
-                                               vector<string> comments_or_null) {
-        if (description.empty()) {
+                                               origin_type org_type, string resource, vector<string> comments) :
+        _description(move(description)), _line_number(line_number), _end_line_number(end_line_number),
+        _origin_type(org_type), _resource_or_null(move(resource)), _comments_or_null(move(comments))
```
MikaelSmith: Does it still make sense to call this _resource_or_null?

Magisus: I was actually curious about that. They won't be null per se, but they will be empty sometimes. Do we care to make that distinction somehow? If not, there's no reason for a special name. Though the empty check will probably still need to happen, like the null check in Java, I assume (haven't found where these are used yet).

MikaelSmith: For now I would assume an empty string means it wasn't initialized.

MikaelSmith: Can leave it for now, revisit later.

MikaelSmith: We'll have to test out cases like that. boost::optional is one way to deal with that (or using unique pointers).

PR #2, opened 2015-08-04
(HC-4) Port the Tokenizer class

Magisus:
This still needs tests, but I'm putting up the main class now, in case anyone wants to review, since it's a lot of code.

Tests on the way!


Magisus:
So, apparently GCC doesn't properly implement the move constructor for streams, despite what the reference says (and this compiles with Clang). These streams are configured somewhere else entirely (see https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/impl/Parseable.java) and then passed to where they're needed in the original project. How do we want to handle this?

MikaelSmith:
Needs a rebase now.

MikaelSmith:
Still writing tests for this?

lib/src/tokenizer.cc
```diff
@@ -0,0 +1,555 @@
...
+            _line_origin = _origin.with_line_number(_line_number);
+            return newline;
+        } else {
+            shared_ptr<token> t;
+            if (start_of_comment(c)) {
+                t = pull_comment(c);
+            } else {
+                switch (c) {
+                    case '"':
+                        t = pull_quoted_string();
```
MikaelSmith: I wish switch had a more succinct syntax; it always makes me want to use a unordered_map instead, even though there's no good reason to.

Magisus: I know that feel, I hate all those breaks. Adds so many lines. But I think this is probably the easiest to read...

Magisus:
Yes, more tests incoming. Also, interesting bug: they have a test for tokenizing the null character, which is supposed to return basically an empty string token. But the C++ is reading it as a terminating character and erroring... Is there a way to handle this to get the expected behavior?

lib/src/tokenizer.cc
```diff
@@ -0,0 +1,555 @@
...
+                    int c = next_char_raw();
+                    if (c == -1) {
+                        throw config_exception("End of input but expecting 4 hex digits for \\uXXXX escape");
+                    }
+                    utf[i] = c;
+                }
+                original += string(utf);
+                short character;
+                sscanf(utf, "%hx", &character);
+                wchar_t buffer[] { static_cast<wchar_t>(character), '\0'};
```
MikaelSmith: Try converting buffer to a wstring with a length of 1. boost::nowide::narrow is probably doing strlen, which will see a length of 0 because we're dealing with the null byte.

lib/src/tokenizer.cc
kylog: In other projects with massive mechanical switch statements, I've seen these three lines collapsed into one, perhaps with whitespace to make it columnar, e.g.

```
case '"':   t = pull_quoted_string();   break;
case '$':   t = pull_substitution();    break;
...
```

It approaches being a wall-of-text, but for a longer switch statement (this may not qualify) that can mean you can fit it onto one page more easily.

Magisus:
Alright, that should do it for tests.

lib/tests/tokenizer_test.cc
```diff
@@ -0,0 +1,694 @@
...
+                line_token(1),
+                tokens::end_token()
+        };
+        tokenizer_test(source, expected);
+    }
+}
+
+TEST_CASE("escape sequence", "[tokenizer]") {
+    SECTION("unicode infinity symbol") {
+        string source = "\"\u221E\"";
```
MikaelSmith: Is this the right number of escape characters?

lib/src/tokenizer.cc
MikaelSmith: Did this suggestion not work?

Magisus: We didn't need it. The error was on the other end, on the input string. My typo. It works fine as is.

lib/src/config_util.cc
```diff
@@ -0,0 +1,29 @@
+#include <internal/config_util.hpp>
+#include <cctype>
+
+namespace hocon {
+
+    bool is_whitespace(char c) {
+        switch (c) {
+            case ' ':
+            case '\n':
+            // TODO: these complain about char overflow, what do.
```
whopper: Hah, are we leaving these cases for future investigation?

lib/src/tokens.cc
```diff
@@ -0,0 +1,223 @@
...
+
+    shared_ptr<simple_config_origin> value::origin() const {
+        return _value->origin();
+    }
+
+    bool value::operator==(const token& other) const {
+        return other.get_token_type() == token_type::VALUE &&
+               other.to_string() == to_string();
+    }
+
```
whopper: Super nitpick: Potentially one too many newlines here

lib/src/config_util.cc
Magisus: They represent various kinds of non-breaking spaces not supported in Java's or C++'s whitespace checking methods. In order to compare against them, we probably would have to cast `c` to one of the bigger char types, but that doesn't make much sense to me, because if `c` couldn't hold those values when it came in, that means it will never equal one of them... Unless I'm understanding this wrong.

lib/tests/tokenizer_test.cc
```diff
@@ -0,0 +1,694 @@
...
+    while(iter.has_next()) {
+        auto size = tokens.size();
+        tokens.push_back(iter.next());
+    }
+    return tokens;
+}
+
+void tokenizer_test(string source, vector<shared_ptr<token>> expected) {
+    vector<shared_ptr<token>> result = tokenize_as_list(source);
+    //REQUIRE(expected.size() == result.size());
```
whopper: Should this be removed?

lib/src/config_util.cc
MikaelSmith: That's true. It means passing around single characters may not work.

Magisus: The original library makes the assumption that *any* character could come up. Do we need that kind of support for our use cases?

MikaelSmith: Probably; we need to support the same input. Mostly that doesn't require extra work on our part - we just read them into a string as bytes - but if we're manipulating or comparing against the strings we need to do extra work.

MikaelSmith: Places I suggest starting to read up on are [ICU4C](http://site.icu-project.org/). Also, Java uses UTF-16; we need to watch for assumptions the Java code might make related to that.

lib/CMakeLists.txt
```diff
@@ -18,10 +18,15 @@ set(PROJECT_SOURCES
     src/values/abstract_config_value.cc
     src/values/config_boolean.cc
     src/values/config_number.cc
-    src/values/config_int.cc
```
MikaelSmith: What happened to config_int? All header-only?

lib/inc/internal/tokenizer.hpp
```diff
@@ -0,0 +1,98 @@
...
+    class problem_exception : std::runtime_error {
+    public:
+        problem_exception(problem prob);
+        problem const& get_problem() const;
+
+    private:
+        problem _problem;
+    };
+
+    class config_exception : public std::runtime_error {
```
whopper: Looking at the ruby port, it looks like we might want `config_exception` available everywhere so we can extend it with things like `config_bug_or_broken_error`. Do you think we should break this out into it's own `config_error.hpp`? I actually have need of it in the `abstract_config_node` class too. (I could also be totally wrong and we're doing it some other way)

lib/src/config_util.cc
peterhuene: Perhaps use `u8_to_u32_iterator` (http://www.boost.org/doc/libs/1_58_0/libs/regex/doc/html/boost_regex/ref/internal_details/uni_iter.html) to enumerate as u32 characters?

lib/inc/internal/tokenizer.hpp
Magisus: No it's fine to move it, I just put it in the only place it'd come up so far. The exception handling in this class is also most likely not in its final form yet, because I didn't understand what they were doing with it. The tests gave some hint, but I want to see it in action elsewhere before I switch everything over to using problem_exception, since I think the implementation will be a bit different than the Java.

lib/src/config_util.cc
MikaelSmith: It might be simpler to build this around regex.

I'm ok recording this as a data point to fix up later, as we get a better sense of what we need to do for unicode in the project. I'd file a ticket to track it.

lib/CMakeLists.txt
Magisus: In the Java code they have both ConfigInt and ConfigLong. ConfigInt returns its value as a long, unless shoved through ConfigNumber's intValueRangeChecked() method. The Ruby only has config_int, because of the way numbers are handled in Ruby. I decided to only implement config_long, because almost everywhere, they seem to just assume the value is a long anyway. 

However, not opposed to re-adding config_int.

lib/src/config_util.cc
Magisus: Sounds good, I'll do that.

lib/CMakeLists.txt
Magisus: Changed my mind on further inspection. We now make the distinction between ints and longs, choosing ints whenever possible based on number size, including when a whole-number double is passed. This reflects what was done in the Java.

Magisus: One reason for my original decision was that the Java's equality testing for config_nubmers deliberately ignores the distinction. This means that the tests I wrote still pass, despite the fact I created config_longs everywhere, and they all should be ints now. I guess this isn't a problem?

lib/inc/internal/tokens.hpp
```diff
@@ -0,0 +1,131 @@
...
+        bool operator==(const token& other) const override;
+    };
+
+    class ignored_whitespace : public token {
+    public:
+        ignored_whitespace(std::shared_ptr<simple_config_origin> origin, std::string whitespace);
+
+        std::string to_string() const override;
+
+        bool operator==(const token& other) const;
```
MikaelSmith: Should this use the override keyword?

lib/src/token.cc
```diff
@@ -0,0 +1,48 @@
...
+
+    int token::line_number() const {
+        if (_origin) {
+            return _origin->line_number();
+        } else {
+            return -1;
+        }
+    }
+
+    shared_ptr<simple_config_origin> token::origin() const {
```
MikaelSmith: We should return this by const&. We're granting access to an object owned by token, but we don't want to make a copy of the shared_ptr (which is more expensive than returning a reference) unless someone else also wants to take ownership of it. So if we return a const& here, others can use it, and take by value when they need to copy it.

lib/src/tokens.cc
```diff
@@ -0,0 +1,222 @@
...
+        return "'${" + result + "}'";
+    }
+
+    bool substitution::operator==(const token& other) const {
+        return other.get_token_type() == token_type::SUBSTITUTION &&
+               to_string() == other.to_string();
+    }
+
+    /** Singleton tokens */
+    shared_ptr<token> tokens::start_token() {
```
MikaelSmith: All of these helpers (and any functions that directly return the result of calling them) should return const& shared_ptrs, so we avoid reference counting unless the caller needs to claim ownership of the token as well.

lib/src/tokenizer.cc
```diff
@@ -0,0 +1,556 @@
...
+                                               original);
+    }
+
+    shared_ptr<token> token_iterator::pull_plus_equals() {
+        int c = next_char_raw();
+        if (c != '=') {
+            throw config_exception("'+' not followed by '=', '" + string(1, c) +
+                "' not allowed after '+'");
+        }
+        return tokens::plus_equals_token();
```
MikaelSmith: Once plus_equals_token is updated, return const& shared_ptr.

lib/src/tokenizer.cc
```diff
@@ -0,0 +1,556 @@
...
+            }
+        } while (true);
+
+        return make_shared<substitution>(_line_origin, optional, expression);
+    }
+
+    shared_ptr<token> token_iterator::pull_next_token(whitespace_saver& saver) {
+        char c = next_char_after_whitespace(saver);
+        if (c == -1) {
+            return tokens::end_token();
```
MikaelSmith: This method usually creates a new token, so don't change it to return a const&. That would cause invalid memory use.

lib/src/tokens.cc
```diff
@@ -0,0 +1,222 @@
...
+    value::value(unique_ptr<abstract_config_value> value, string original_text) :
+            token(token_type::VALUE, nullptr, original_text),
+            _value(move(value)) { }
+
+    std::string value::to_string() const {
+        return _value->transform_to_string();
+    }
+
+    shared_ptr<simple_config_origin> value::origin() const {
+        return _value->origin();
```
MikaelSmith: Once _value->origin() returns const&, so can this function.

lib/src/values/abstract_config_value.cc
```diff
@@ -13,4 +13,12 @@ namespace hocon {
         return "";
     }
 
+    shared_ptr<simple_config_origin> abstract_config_value::origin() const {
```
MikaelSmith: Can return a const& shared_ptr.

lib/src/values/config_number.cc
```diff
@@ -31,4 +37,32 @@ namespace hocon {
...
+        if (l < numeric_limits<int>::min() || l > numeric_limits<int>::max()) {
+            throw config_exception("Tried to get int from out of range value " + to_string(l));
+        }
+        return static_cast<int>(l);
+    }
+
+    unique_ptr<config_number> config_number::new_number(
+            shared_ptr<simple_config_origin> origin, int64_t value, std::string original_text) {
+        if (value >= numeric_limits<int>::min() && value <= numeric_limits<int>::max()) {
+            return unique_ptr<config_int>(new config_int(origin, static_cast<int>(value), original_text));
```
MikaelSmith: These constructor calls should do `move(origin)`.

lib/src/values/config_number.cc
```diff
@@ -31,4 +37,32 @@ namespace hocon {
...
+        } else {
+            return unique_ptr<config_long>(new config_long(origin, value, original_text));
+        }
+    }
+
+    unique_ptr<config_number> config_number::new_number(
+            shared_ptr<simple_config_origin> origin, double value, std::string original_text) {
+        int64_t as_long = static_cast<int64_t>(value);
+        if (as_long == value) {
+            return new_number(origin, as_long, original_text);
```
MikaelSmith: Should 'move(origin)'.

lib/src/values/config_string.cc
```diff
@@ -0,0 +1,25 @@
+#include <internal/values/config_string.hpp>
+
+using namespace std;
+
+namespace hocon {
+
+    config_string::config_string(shared_ptr<simple_config_origin> origin, string text, config_string_type quoted) :
+        abstract_config_value(origin), _text(move(text)), _quoted(quoted) { }
```
MikaelSmith: Should `move(origin)`.

lib/src/values/abstract_config_value.cc
```diff
@@ -1,11 +1,11 @@
...
+#include <internal/values/abstract_config_value.hpp>
 
 using namespace std;
 
 namespace hocon {
 
-    abstract_config_value::abstract_config_value(simple_config_origin origin) :
-        _origin(move(origin))
+    abstract_config_value::abstract_config_value(shared_ptr<simple_config_origin> origin) :
+        _origin(origin)
```
MikaelSmith: `move(origin)`.

lib/src/values/config_boolean.cc
```diff
@@ -1,11 +1,11 @@
...
+#include <internal/values/config_boolean.hpp>
 
 using namespace std;
 
 namespace hocon {
 
-    config_boolean::config_boolean(simple_config_origin origin, bool value) :
-        abstract_config_value(move(origin)), _value(value) { }
+    config_boolean::config_boolean(shared_ptr<simple_config_origin> origin, bool value) :
+        abstract_config_value(origin), _value(value) { }
```
MikaelSmith: `move(origin)`

lib/src/values/config_null.cc
```diff
@@ -1,11 +1,11 @@
...
+#include <internal/values/config_null.hpp>
 
 using namespace std;
 
 namespace hocon {
 
-    config_null::config_null(simple_config_origin origin) :
-            abstract_config_value(move(origin)) { }
+    config_null::config_null(shared_ptr<simple_config_origin> origin) :
+            abstract_config_value(origin) { }
```
MikaelSmith: `move(origin)`

lib/src/values/config_number.cc
```diff
@@ -1,10 +1,16 @@
...
+#include <limits>
 
 using namespace std;
 
 namespace hocon {
 
-    config_number::config_number(simple_config_origin origin, string original_text) :
+    config_number::config_number(shared_ptr<simple_config_origin> origin,
+                                 string original_text) :
             abstract_config_value(origin), _original_text(original_text) { }
```
MikaelSmith: Use `move` on origin and original_text.

lib/src/values/config_number.cc
```diff
@@ -31,4 +37,32 @@ namespace hocon {
     bool config_number::operator!=(const config_number &other) const {
         return !(*this == other);
     }
+
+    int config_number::int_value_range_checked(std::string path) {
```
MikaelSmith: What's this used for, why does it take a path argument, and path should be passed by const&.

lib/src/values/config_number.cc
Magisus: So I have a question about this. If you do move(origin), doesn't it null out the original? Often the same shared_ptr<origin> is used to create multiple tokens. Will that cause a problem if we use move, or am I understanding it wrong?

lib/src/values/config_int.cc
```diff
@@ -1,11 +1,11 @@
...
+#include <internal/values/config_int.hpp>
 
 using namespace std;
 
 namespace hocon {
 
-    config_int::config_int(simple_config_origin origin, int value, string original_text) :
-            config_number(move(origin), move(original_text)), _value(value) { }
+    config_int::config_int(shared_ptr<simple_config_origin> origin, int value, string original_text) :
+            config_number(origin, move(original_text)), _value(value) { }
```
MikaelSmith: `move(origin)` should still be used

lib/src/values/config_double.cc
```diff
@@ -1,11 +1,12 @@
...
 
 using namespace std;
 
 namespace hocon {
 
-    config_double::config_double(simple_config_origin origin, double value, string original_text) :
-            config_number(move(origin), move(original_text)), _value(value) { }
+    config_double::config_double(shared_ptr<simple_config_origin> origin,
+                                 double value, string original_text) :
+            config_number(origin, move(original_text)), _value(value) { }
```
MikaelSmith: Should `move(origin)`, as it's still cheaper than copying a shared_ptr.

MikaelSmith:
Discussion on moving shared_ptr: http://stackoverflow.com/questions/10953325/should-i-stdmove-a-shared-ptr-in-a-move-constructor

lib/src/values/config_number.cc
Magisus: This is used for something way down the line. As I pointed out in my comments on why I originally removed config_int, config_int only returns its value as a long (to help make them more interchangeable, I guess). There are a couple of places where the Java decides to make a distinction between longs and ints, and this method is used to get the int value if possible. It's kinda weird, I know. As far as I can tell, the path has something to do with where in the document structure the number sits, but I honestly don't know how it works yet.

lib/src/values/config_number.cc
MikaelSmith: The function new_number is taking origin by value, so it's already created a copy, and invalidating that copy will only affect other attempts to use it in the local scope (which there won't be, since you return after the move). By doing a move on shared_ptr, you're only moving the shared_ptr object, not the simple_config_origin it refers to.

lib/src/values/config_number.cc
MikaelSmith: Ok. Then it should still probably take path by const&.

lib/src/tokenizer.cc
```diff
@@ -0,0 +1,556 @@
...
+        }
+    }
+
+    /**
+     * Called if the next token is not a simple value;
+     * discards any whitespace we were saving between
+     * simple values.
+     */
+    shared_ptr<token> token_iterator::whitespace_saver::next_is_not_simple_value(
+            simple_config_origin base_origin, int line_number)
```
MikaelSmith: This could probably take by const&, since it's just going to use it to create a copy (not take ownership of it).

lib/src/tokenizer.cc
```diff
@@ -0,0 +1,556 @@
...
+
+    /** Whitespace Saver */
+    token_iterator::whitespace_saver::whitespace_saver() : _last_token_was_simple_value(false) { }
+
+    void token_iterator::whitespace_saver::add(char c) {
+        _whitespace += c;
+    }
+
+    shared_ptr<token> token_iterator::whitespace_saver::check(token_type type,
+                                                              simple_config_origin base_origin,
```
MikaelSmith: This should take `base_origin` by const&.

lib/src/tokenizer.cc
```diff
@@ -0,0 +1,556 @@
...
+        return create_whitespace_token(base_origin, line_number);
+    }
+
+    /**
+     * Called if the next token IS a simple value,
+     * so creates a whitespace token if the previous
+     * token also was simple.
+     */
+    shared_ptr<token> token_iterator::whitespace_saver::next_is_simple_value(
+            simple_config_origin base_origin, int line_number) {
```
MikaelSmith: `base_origin` by const&

lib/src/tokenizer.cc
```diff
@@ -0,0 +1,556 @@
...
+        shared_ptr<token> t = create_whitespace_token(base_origin, line_number);
+        if (!_last_token_was_simple_value) {
+            _last_token_was_simple_value = true;
+        }
+        return t;
+    }
+
+    shared_ptr<token> token_iterator::whitespace_saver::create_whitespace_token(
+            simple_config_origin base_origin, int line_number) {
```
MikaelSmith: `base_origin` by const&

lib/src/tokenizer.cc
```diff
@@ -0,0 +1,556 @@
...
+        return c;
+    }
+
+    bool token_iterator::is_simple_value(token_type type) {
+        return type == token_type::SUBSTITUTION ||
+                type == token_type::VALUE ||
+                type == token_type::UNQUOTED_TEXT;
+    }
+
+    shared_ptr<simple_config_origin> token_iterator::line_origin(simple_config_origin origin,
```
MikaelSmith: `origin` by const&

lib/inc/internal/simple_config_origin.hpp
```diff
@@ -16,6 +17,14 @@ namespace hocon {
         simple_config_origin(std::string description, int line_number = -1, int end_line_number = -1,
                              origin_type org_type = origin_type::GENERIC);
 
+        int line_number();
+
+        /**
+         * Returns a pointer to a copy of this origin with the specified line number
+         * as both starting and ending line.
+         */
+        std::shared_ptr<simple_config_origin> with_line_number(int line_number);
```
MikaelSmith: This should be a const operation.

lib/inc/internal/simple_config_origin.hpp
```diff
@@ -16,6 +17,14 @@ namespace hocon {
         simple_config_origin(std::string description, int line_number = -1, int end_line_number = -1,
                              origin_type org_type = origin_type::GENERIC);
 
+        int line_number();
```
MikaelSmith: This should be a const operation.

lib/tests/test_utils.cc
```diff
@@ -0,0 +1,56 @@
...
+
+using namespace std;
+
+namespace hocon {
+
+    shared_ptr<simple_config_origin> fake_origin(string description, int line_number) {
+        return make_shared<simple_config_origin>(move(description), line_number, line_number, origin_type::GENERIC);
+    }
+
+    shared_ptr<value> string_token(string const& text, config_string_type type) {
```
MikaelSmith: This is definitely not significant, but normally it's better to take string by value in a constructor if you plan to take ownership of it (as config_string does, which you'd need to do move(text) to). You do that most places, so this looks like an oversight.

lib/src/tokenizer.cc
MikaelSmith: Another newer approach is using lambda functions and return, as in
```
[c]() {
    switch (c) {
        case '"': return pull_quoted_string();
        case '$': return pull_substitution();
        default: return nullptr;
    }
}();
```

I don't think it's important to change it however.

PR #3, opened 2015-08-12
(HC-14) and (HC-15)

Magisus:
(HC-14) Port ConfigNode parent classes

Adds generic classes necessary to port the rest of the CofigNode specific implementation classes.
Adpated from initial investigation code written by William Hopper.

(HC-15) Port ConfigNodeInclude, ConfigNodeSimpleValue, ConfigNodeSingleToken, and ConfigNodeComment

Adds some basic implementations of abstract_config_node and their associated tests.


MikaelSmith:
cpplint's unhappy

lib/src/nodes/config_node_single_token.cc
```diff
@@ -0,0 +1,18 @@
...
+
+using namespace std;
+
+namespace hocon {
+
+    config_node_single_token::config_node_single_token(shared_ptr<token> t) :
+            _token(move(t)) { }
+
+    vector<shared_ptr<token>> config_node_single_token::get_tokens() const {
+        return vector<shared_ptr<token>> { _token };
```
MikaelSmith: I think you could safely `return { _token };`.

lib/inc/internal/token.hpp
```diff
@@ -6,6 +6,7 @@
 
 namespace hocon {
 
+
```
MikaelSmith: Pedantry: extra space.

lib/inc/internal/nodes/config_node_simple_value.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <internal/values/abstract_config_value.hpp>
+
+namespace hocon {
+
+    class config_node_simple_value : public abstract_config_node {
+    public:
+        config_node_simple_value(shared_token value);
+
+        shared_token get_token() const;
+        std::unique_ptr<abstract_config_value> const& get_value() const;
```
MikaelSmith: With unique_ptr, it makes more sense to just return a const& of the object. unique_ptr isn't copyable, so there's no useful reason to have access to a `const&` of it.

Magisus: Well some kind of pointer is necessary, because abstract type. You can't declare an abstract return type. How should this be handled?

lib/src/nodes/config_node_simple_value.cc
```diff
@@ -0,0 +1,44 @@
...
+    shared_token config_node_simple_value::get_token() const {
+        return _value;
+    }
+
+    token_list config_node_simple_value::get_tokens() const {
+        return token_list { _value };
+    }
+
+    unique_ptr<abstract_config_value> const& config_node_simple_value::get_value() const {
+        shared_ptr<value> value_token = dynamic_pointer_cast<value>(_value);
```
MikaelSmith: I had no idea `dynamic_pointer_cast` existed.

Magisus: I think I actually found it thanks to CLion suggesting methods. I typed "dynamic" and saw it, then googled it because I was curious. Seemed to be exactly what I wanted.

lib/src/nodes/config_node_simple_value.cc
```diff
@@ -0,0 +1,44 @@
...
+        if (value_token) {
+            return value_token->get_value();
+        }
+
+        // TODO: We have two separate cases here: in the above, we have an existing value
```
MikaelSmith: We may need to use shared_ptr everywhere for wrapping abstract_config_value types.

lib/inc/internal/nodes/config_node_simple_value.hpp
MikaelSmith: You should be able to return an abstract reference. We probably should look more closely at how config_values are being used, it would make sense to use shared_ptr everywhere if they're expected to be owned by multiple objects.

lib/src/nodes/config_node_include.cc
```diff
@@ -0,0 +1,40 @@
...
+namespace hocon {
+
+    config_node_include::config_node_include(vector<unique_ptr<abstract_config_node>> children,
+                                             config_include_kind kind) :
+        _children(move(children)), _kind(kind) { }
+
+    token_list config_node_include::get_tokens() const {
+        token_list tokens;
+        for (auto&& node : _children) {
+            for (auto&& token : node->get_tokens()) {
```
MikaelSmith: Instead of looping over get_tokens, I think you could do `tokens.insert(tokens.end(), node->get_tokens().begin(), node->get_tokens().end())`. I wish that weren't longer, because conceptually it's simpler and performs better.

lib/src/nodes/config_node_include.cc
```diff
@@ -0,0 +1,40 @@
...
+        return _children;
+    }
+
+    config_include_kind config_node_include::kind() const {
+        return _kind;
+    }
+
+    string config_node_include::name() const {
+        for (auto&& node : _children) {
+            config_node_simple_value* simple = dynamic_cast<config_node_simple_value*>(node.get());
```
MikaelSmith: It's not a common pattern, but you can declare a variable and check if it's null in the expression of an if block. So
```
if (config_node_simple_value* simple = dynamic_cast<config_node_simple_value*>(node.get())) {
    return simple->get_value()->transform_to_string();
}
```
should work.

lib/inc/internal/token.hpp
```diff
@@ -39,4 +39,7 @@ namespace hocon {
         std::string _debug_string;
     };
 
+    using token_list = std::vector<std::shared_ptr<token>>;
+    using shared_token = std::shared_ptr<token>;
```
MikaelSmith: I'd make this `shared_ptr<const token>`, and declare `token_list` in terms of `vector<shared_token>`.

lib/inc/internal/values/abstract_config_value.hpp
```diff
@@ -21,4 +20,6 @@ namespace hocon {
         std::shared_ptr<simple_config_origin> _origin;
     };
 
+    using shared_value = std::shared_ptr<abstract_config_value>;
```
MikaelSmith: Make `shared_ptr<const abstract_config_value>`?

lib/src/nodes/config_node_simple_value.cc
```diff
@@ -0,0 +1,41 @@
...
+#include <internal/values/config_string.hpp>
+
+using namespace std;
+
+namespace hocon {
+
+    config_node_simple_value::config_node_simple_value(shared_token value) : _value(move(value)) { }
+
+    shared_token config_node_simple_value::get_token() const {
+        return _value;
```
MikaelSmith: It might be clearer to call it `_token` instead of `_value`. That naming is throwing me off, and doesn't seem to match the java source.

MikaelSmith:
I think those are the last fixes.

MikaelSmith:
Sometime it might make sense to make `unique_ptr<abstract_config_node>` be `unique_ptr<const abstract_config_node>`, but it shouldn't cause any problems yet.

PR #5, opened 2015-08-14
(HC-16) Port ConfigNodePath and needed Path, PathBuilder and PathParser classes

Magisus:
Adds path functionality and tests.

### DIFF
--- a/lib/inc/internal/nodes/config_node_path.hpp
+++ b/lib/inc/internal/nodes/config_node_path.hpp
@@ -1,0 +1,22 @@
+#pragma once
+
+#include <internal/nodes/abstract_config_node.hpp>
+#include <internal/path.hpp>
+#include <internal/token.hpp>
+
+namespace hocon {
+
+    class config_node_path : public abstract_config_node {
+    public:
+        config_node_path(path node_path, token_list tokens);
+
+        token_list get_tokens() const override;
+        path get_path() const;
+
+
+    private:
+        path _path;
+        token_list _tokens;
+    };
+
+}  // namespace hocon
--- a/lib/inc/internal/path_parser.hpp
+++ b/lib/inc/internal/path_parser.hpp
@@ -1,0 +1,55 @@
+#pragma once
+
+#include "path_builder.hpp"
+#include "simple_config_origin.hpp"
+#include "tokenizer.hpp"
+#include <hocon/config_syntax.hpp>
+#include <internal/nodes/config_node_path.hpp>
+
+#include <string>
+#include <vector>
+
+namespace hocon  {
+
+    class path_parser {
+    public:
+        static config_node_path parse_path_node(std::string const& path_string,
+                                                config_syntax flavor = config_syntax::CONF);
+
+        static path parse_path(std::string const& path_string);
+
+        static path parse_path_expression(token_iterator& expression, simple_config_origin origin,
+                                          std::string const& original_text = "",
+                                          token_list* path_tokens = nullptr,
+                                          config_syntax flavor = config_syntax::CONF);
+
+        static config_node_path parse_path_node_expression(token_iterator& expression,
+                                                           simple_config_origin origin,
+                                                           std::string const& original_text = "",
+                                                           config_syntax flavor = config_syntax::CONF);
+
+    private:
+        class element {
+        public:
+            element(std::string initial, bool can_be_empty);
+
+            std::string to_string() const;
+
+            std::string _value;
+            bool _can_be_empty;
+        };
+
+        static token_list split_token_on_period(shared_token t, config_syntax flavor);
+
+        static void add_path_text(std::vector<element>& buff, bool was_quoted, std::string new_text);
+
+        static bool looks_unsafe_for_fast_parser(std::string s);
+
+        static path fast_path_build(path tail, std::string s);
+
+        static path speculative_fast_parse_path(std::string const& path);
+
+        static const simple_config_origin api_origin;
+    };
+
+}  // namespace hocon
--- a/lib/src/config_util.cc
+++ b/lib/src/config_util.cc
@@ -1,5 +1,6 @@
 #include <internal/config_util.hpp>
-#include <cctype>
+
+using namespace std;
 
 namespace hocon {
 
@@ -24,6 +25,41 @@ namespace hocon {
 
     bool is_C0_control(char c) {
         return c >= 0 && c <= 0x001F;
+    }
+
+    string render_json_string(string const& s) {
+        string result = "\"";
+        for (char c : s) {
+            switch (c) {
+                case '"':
+                    result += "\\\"";
+                    break;
+                case '\\':
+                    result += "\\\\";
+                    break;
+                case '\n':
+                    result += "\\n";
+                    break;
+                case '\b':
+                    result += "\\b";
+                    break;
+                case '\f':
+                    result += "\\f";
+                    break;
+                case '\r':
+                    result += "\\r";
+                    break;
+                case '\t':
+                    result += "\\t";
+                    break;
+                default:
+                    // TODO: The java has a case here for is_C0_control() that converts c to
+                    // a unicode literal (\uXXXX). Not sure how to handle at present.
+                    result += c;
+            }
+        }
+        result += "\"";
+        return result;
     }
 
 }  // namespace hocon
--- a/lib/src/nodes/config_node_path.cc
+++ b/lib/src/nodes/config_node_path.cc
@@ -1,0 +1,17 @@
+#include <internal/nodes/config_node_path.hpp>
+
+using namespace std;
+
+namespace hocon {
+
+    config_node_path::config_node_path(path node_path, token_list tokens) :
+        _path(move(node_path)), _tokens(move(tokens)) { }
+
+    token_list config_node_path::get_tokens() const {
+        return _tokens;
+    }
+
+    path config_node_path::get_path() const {
+        return _path;
+    }
+}  // namespace hocon
--- a/lib/src/path.cc
+++ b/lib/src/path.cc
@@ -1,0 +1,211 @@
+#include <internal/path.hpp>
+#include <internal/config_exception.hpp>
+#include <internal/path_builder.hpp>
+#include <internal/config_util.hpp>
+#include <internal/path_parser.hpp>
+
+using namespace std;
+
+namespace hocon {
+
+    path::path() { }
+
+    path::path(List<shared_string> other_path) : _path(move(other_path)) { }
+
+    path::path(string first, path const& remainder) :
+            _path(make_shared<string>(move(first)), remainder._path) { }
+
+    path::path(vector<string> elements) {
+        if (elements.size() == 0) {
+            throw config_exception("Empty path");
+        }
+
+        path remainder;
+        if (elements.size() > 1) {
+            path_builder builder;
+            for (size_t i = 1; i < elements.size(); i++) {
+                builder.append_key(elements[i]);
+            }
+            remainder = builder.result();
+        }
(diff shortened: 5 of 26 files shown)


lib/tests/tokenizer_test.cc
```diff
@@ -118,7 +118,7 @@ TEST_CASE("tokenize all types", "[tokenizer]") {
                 unquoted_text_token(" "),
                 substitution_token(string_token("x.y", config_string_type::UNQUOTED), true),
```
MikaelSmith: Should this be changed to unquoted_text_token, or is it unchanged on purpose?

MikaelSmith:
Can you move functional_list.hpp into Leatherman as a vendor library (similar to catch), and note the source at https://github.com/BartoszMilewski/Okasaki/tree/master/List. It also has a test case we could bring along.

lib/src/config_util.cc
```diff
@@ -26,4 +27,40 @@ namespace hocon {
         return c >= 0 && c <= 0x001F;
     }
 
+    string render_json_string(string s) {
+        string result = "\"";
+        for (size_t i = 0; i < s.length(); i++) {
```
MikaelSmith: This could be a range-based for loop.

lib/src/config_util.cc
```diff
@@ -26,4 +27,40 @@ namespace hocon {
         return c >= 0 && c <= 0x001F;
     }
 
+    string render_json_string(string s) {
```
MikaelSmith: String argument should be a const&, no need to copy it.

MikaelSmith:
@Magisus Changed my mind, we can hold off on moving it to Leatherman until we get the parser working with it and decide we'll keep it.

lib/inc/internal/nodes/config_node_complex_value.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include "abstract_config_node_value.hpp"
+
+namespace hocon {
+
+    class config_node_complex_value : public abstract_config_node_value {
```
MikaelSmith: Does this need implementation? I don't see any of the methods defined.

lib/src/path.cc
```diff
@@ -0,0 +1,219 @@
...
+        if (paths_to_concat.size() == 0) {
+            throw config_exception("Empty path");
+        }
+
+        path_builder builder;
+        if (paths_to_concat[0].has_remainder()) {
+            builder.append_path(paths_to_concat[0].remainder());
+        }
+
+        for (size_t i = 1; i < paths_to_concat.size(); i++) {
```
MikaelSmith: Use range-based for-loop over paths_to_concat?

lib/inc/internal/nodes/config_node_complex_value.hpp
Magisus: Whoops, I did some rebasing and stuff to isolate the path code, forgot to remove this file. It's written but the tests still need things that haven't been implemented yet, so I decided not to put it up yet.

lib/src/path.cc
Magisus: Can you start at some other element besides the first one? This loop needs to start at index 1.

Magisus: Looking at the solutions people have posted for doing that, I'm in favor of just keeping this a regular for loop.

MikaelSmith: The one I've done before is `for (auto &&p : boost::make_iterator_range(paths_to_concat.begin()+1, paths_to_concat.end())) { ... }`. I'm ok leaving it for now.

lib/src/path.cc
```diff
@@ -0,0 +1,219 @@
...
+    bool path::operator!=(path const& other) const {
+        return !(*this == other);
+    }
+
+    bool path::has_funky_chars(string const& s) const {
+        if (s.length() == 0) {
+            return false;
+        }
+
+        for (size_t i = 0; i < s.length(); i++) {
```
MikaelSmith: This looks like it could definitely be a range-based for-loop.

lib/inc/internal/nodes/config_node_path.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include <internal/path.hpp>
+#include <internal/token.hpp>
+
+namespace hocon {
+
+    class config_node_path : public abstract_config_node {
+    public:
+        config_node_path(path node_path, token_list tokens);
+
+        token_list get_tokens() const;
```
MikaelSmith: Should include `override`.

lib/inc/internal/path_parser.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <internal/nodes/config_node_path.hpp>
+
+#include <string>
+#include <vector>
+
+namespace hocon  {
+
+    class path_parser {
+    public:
+        static config_node_path parse_path_node(std::string const& path_string);
```
MikaelSmith: Any reason you chose 2 methods over using a default argument?

lib/src/nodes/config_node_path.cc
```diff
@@ -0,0 +1,17 @@
+#include <internal/nodes/config_node_path.hpp>
+
+using namespace std;
+
+namespace hocon {
+
+    config_node_path::config_node_path(path node_path, token_list tokens) :
+        _path(move(node_path)), _tokens(move(tokens)) { }
+
+    token_list config_node_path::get_tokens() const {
```
MikaelSmith: This seems like it could return a `token_list const&`, to avoid copying the vector if we just need access to the contents.

MikaelSmith:
I think this is ok besides my comments above. I'd personally choose more range-based for-loops, but I understand keeping the original Java loops for simplicity.

lib/src/nodes/config_node_path.cc
Magisus: In config_node_include, we are creating a list of tokens in this method (this inherits from abstract_config_node). As is, we can't return a reference from that. Suggestions for compromise?

MikaelSmith: We could stash the created list in config_node_include and return a reference to it. Depends on how the result of get_tokens is used; if we always want a copy to manipulate then stashing a copy in the object isn't great.

Let's leave it as-is, and we can optimize later if it's an issue.

lib/inc/internal/path_parser.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+        static void add_path_text(std::vector<element>& buff, bool was_quoted, std::string new_text);
+
+        static bool looks_unsafe_for_fast_parser(std::string s);
+
+        static path fast_path_build(path tail, std::string s);
+
+        static path speculative_fast_parse_path(std::string const& path);
+
+        static simple_config_origin api_origin;
```
MikaelSmith: This should probably be const.

lib/src/path_parser.cc
```diff
@@ -0,0 +1,251 @@
...
+                    // separators, because we basically want "foo 3.0bar" to parse as a string even
+                    // though there's a number in it. The fact that we tokenize non-string values
+                    // is largely an implementation detail.
+                    auto value = tokens::get_value(t);
+
+                    // We need to split the tokens on a . so that we can get sub-paths but still preserve
+                    // the original path text when doing an insertion
+                    if (path_tokens != nullptr) {
+                        path_tokens->pop_back();
+                        for (auto&& token : split_token_on_period(t, flavor)) {
```
MikaelSmith: This could use `path_tokens->insert`. The only reason I prefer that is to avoid having to reallocate the token_list when we already know how many elements we're adding.

lib/src/path_parser.cc
```diff
@@ -0,0 +1,251 @@
...
+                            path_tokens->push_back(token);
+                        }
+                    }
+                    text = value->transform_to_string();
+                } else if (t->get_token_type() == token_type::UNQUOTED_TEXT) {
+                    // We need to split the tokens on a . so that we can get sub-paths but still preserve
+                    // the original path text when doing an insertion on ConfigNodeObjects
+                    if (path_tokens != nullptr) {
+                        path_tokens->pop_back();
+                        for (auto&& token : split_token_on_period(t, flavor)) {
```
MikaelSmith: Could use `path_tokens->insert`.

lib/src/path_parser.cc
```diff
@@ -0,0 +1,251 @@
...
+        return builder.result();
+    }
+
+    token_list path_parser::split_token_on_period(shared_token t, config_syntax flavor) {
+        string token_text = t->token_text();
+        if (token_text == ".") {
+            return token_list { t };
+        }
+
+        vector<string> token_strings;
```
MikaelSmith: Sorry you had to write this. I'd do something like https://github.com/puppetlabs/leatherman/blob/master/execution/src/execution.cc#L271-L275 instead. There's also http://www.boost.org/doc/libs/1_58_0/doc/html/string_algo/usage.html#idp424404576, but that creates a copy of each substring. The iterator in execution.cc iterates over token_text itself, presenting each component as a string range.

You should be able to pass string_text directly to make_split_iterator.

MikaelSmith: Sometime we need to create a utility function that wraps that iterator creation, it should be simpler to use.

MikaelSmith: Oh, we just pass it to a constructor. I'd use
```
for (auto &&s : boost::split(token_text, boost::is_any_of("."))) {
    ... make_shared<unquoted_text>(t->origin(), move(s))...
}
```

lib/src/path.cc
```diff
@@ -0,0 +1,215 @@
...
+        return *first() == *other.first() &&
+               remainder() == other.remainder();
+    }
+
+    bool path::operator!=(path const& other) const {
+        return !(*this == other);
+    }
+
+    bool path::has_funky_chars(string const& s) {
+        if (s.length() == 0) {
```
kylog: To me it reads more clearly if this is `if (s.empty()) {` (There might be a performance gain as well, but that's a different conversation.)

If others agree, should the style guide add a preference for using the `empty` method of standard containers over size checks of 0?

Separate question: is this needed here? I.e. will the logic below do the Right Thing with an empty string?

Magisus: Actually, I think you're right, we probably don't need that check now that I've changed how the rest of it works. find_if should just return s.end().

lib/src/path_parser.cc
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    token_list path_parser::split_token_on_period(shared_token t, config_syntax flavor) {
+        string token_text = t->token_text();
+        if (token_text == ".") {
+            return token_list { t };
+        }
+
+        auto token_it = boost::make_iterator_range(make_split_iterator(
+                token_text, token_finder(is_any_of("."), token_compress_on)), split_iterator<string::iterator>());
```
MikaelSmith: I'd recommend `string::const_iterator`. Does `auto &&s : token_it` not work?

lib/inc/internal/path_parser.hpp
MikaelSmith: Is it illegal to declare a static member const and initialize it?

lib/src/path_parser.cc
Magisus: The error seems to be between the return value of make_split_iterator() and the last argument. Either make_split_iterator() has to be made to return a const_iterator somehow, or we can't use const.

lib/inc/internal/path_parser.hpp
Magisus: I think so. It is unhappy when I try to initialize either in the header or in the source.

MikaelSmith:
:+1: squash up

PR #6, opened 2015-08-17
(maint) Move parser headers to hocon public dir

MikaelSmith:
Public hocon header files should be under the hocon directory, so move
parser headers to that directory.


MikaelSmith:
ping @Magisus

Magisus:
Okay, makes sense.

PR #7, opened 2015-08-17
(maint) Add shared_origin alias for simple_config_origin

MikaelSmith:
A few uses of simple_config_origin showed bad practices of creating a
new shared_ptr of a simple_config_origin from a reference drawn from
another shared_ptr. This can lead to freed memory references. The
shared pointers also didn't use const, allowing for the shared versions
of simple_config_origin to be modified.

Change to passing shared_ptr<const simple_config_origin> everywhere,
and add an alias shared_origin to make that simpler.

### DIFF
--- a/lib/inc/internal/path_parser.hpp
+++ b/lib/inc/internal/path_parser.hpp
@@ -18,13 +18,13 @@ namespace hocon  {
 
         static path parse_path(std::string const& path_string);
 
-        static path parse_path_expression(token_iterator& expression, simple_config_origin origin,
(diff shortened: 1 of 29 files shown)


MikaelSmith:
I'll need to rebase on https://github.com/puppetlabs/cpp-hocon/pull/5, and probably change api_token to be a shared_origin.

MikaelSmith:
ping @Magisus, I think this is ready now.

Magisus:
Looks good.

PR #8, opened 2015-08-18
(HC-22) Port ConfigParseOptions

MikaelSmith:
For now we leave config_includer unimplemented, as it's not being
exercised.

config_parse_options was kept an immutable object because we already
wanted to use a shared_ptr for config_includer, to avoid having to make
it a concrete class, and it keeps the implementation more in-line with
the original.

Copies should be light-weight, but we may reconsider and make this a
mutable class to avoid lots of copies if it's never really shared.

### DIFF
--- a/lib/inc/hocon/config_parse_options.hpp
+++ b/lib/inc/hocon/config_parse_options.hpp
@@ -1,0 +1,133 @@
+#pragma once
+
+#include <string>
+#include <memory>
+#include "config_syntax.hpp"
+
+namespace hocon {
+    class config_includer;
+
+    /**
+     * A set of options related to parsing.
+     *
+     * <p>
+     * This object is immutable, so the "setters" returns a new object.
+     *
+     * <p>
+     * Here is an example of creating a custom {@code config_parse_options}:
+     *
+     * <pre>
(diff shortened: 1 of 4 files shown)


MikaelSmith:
No tests to add for this yet, I think ConfigDocument tests should cover it.

MikaelSmith:
ping @Magisus

lib/inc/hocon/config_parse_options.hpp
```diff
@@ -0,0 +1,133 @@
+#pragma once
+
+#include <string>
+#include <memory>
+#include "config_syntax.hpp"
+
+namespace hocon {
+    class config_includer;
```
Magisus: Out of curiosity, what's the benefit of this as opposed to just including the config_includer header? Are there cases where one should be preferred to the other?

MikaelSmith: Using class forward-declaration can cut down on the number of include files introduced to each source file. When config_document needs to use config_parse_options, it doesn't necessarily need to know how config_includer is implemented and can avoid including header files related to it.

It's a way to keep compile time down in C++.

Magisus:
Alright, well this looks good to me. Merging.

PR #10, opened 2015-08-19
HC-17 and HC-18

Magisus:
This is an unfinished implementation of HC-17 and HC-18. All the tests pass except for the last "can replace an object" tests in config_node_test, which segfault. Any help figuring out why one of the post-replacement object's children is null would be appreciated...

### DIFF
--- a/lib/src/nodes/config_node_object.cc
+++ b/lib/src/nodes/config_node_object.cc
@@ -1,0 +1,317 @@
+#include <internal/nodes/config_node_object.hpp>
+#include <internal/nodes/config_node_field.hpp>
+#include <internal/nodes/config_node_single_token.hpp>
+#include <internal/tokens.hpp>
+#include <internal/path_parser.hpp>
+#include <internal/nodes/config_node_include.hpp>
+#include <boost/range/iterator_range_core.hpp>
+
+using namespace std;
+
+namespace hocon {
+
+    config_node_object::config_node_object(shared_node_list children) :
+            config_node_complex_value(move(children)) { }
+
+    shared_ptr<config_node_complex_value> config_node_object::new_node(shared_node_list nodes) {
+        return make_shared<config_node_object>(move(nodes));
+    }
+
+    bool config_node_object::has_value(path desired_path) const {
+        for (auto&& node : children()) {
+            if (auto field = dynamic_pointer_cast<config_node_field>(node)) {
+                path key = field->path()->get_path();
+                if (key == desired_path || key.starts_with(desired_path)) {
+                    return true;
+                } else if (desired_path.starts_with(key)) {
+                    if (auto object = dynamic_pointer_cast<config_node_object>(field->get_value())) {
+                        path remaining_path = desired_path.sub_path(key.length());
+                        if (object->has_value(remaining_path)) {
+                            return true;
+                        }
+                    }
+                }
+            }
+        }
+        return false;
+    }
+
+    shared_ptr<config_node_object> config_node_object::change_value_on_path(path desired_path, shared_node_value value,
+                                                                            config_syntax flavor)
+    {
+        shared_node_list children_copy = children();
+        bool seen_non_matching = false;
+        shared_node_value value_copy = value;
+        for (int i = static_cast<int>(children_copy.size() - 1); i >= 0; i--) {
+            auto&& child = children_copy[i];
+            if (auto single_token = dynamic_pointer_cast<config_node_single_token>(child)) {
+                shared_token t = single_token->get_token();
+                // Ensure that, when we are removing settings in JSON, we don't end up with a trailing comma
+                if (flavor == config_syntax::JSON && !seen_non_matching && t == tokens::comma_token()) {
+                    children_copy.erase(children_copy.begin() + i);
+                }
+                continue;
+            } else if (dynamic_pointer_cast<config_node_field>(child) == nullptr) {
+                continue;
+            }
+            auto field = dynamic_pointer_cast<config_node_field>(child);
+            path key = field->path()->get_path();
+
+            // Delete all multi-element paths that start with the desired path, since technically they are duplicates
+            if ((value_copy == nullptr && key == desired_path) ||
+                    (key.starts_with(desired_path) && !(key == desired_path)))
+            {
+                children_copy.erase(children_copy.begin() + i);
+                //  Remove any whitespace or commas after the deleted setting
+                for (size_t j = i; j < children_copy.size(); j++) {
+                    if (auto single_token = dynamic_pointer_cast<config_node_single_token>(children_copy[j])) {
+                        shared_token t = single_token->get_token();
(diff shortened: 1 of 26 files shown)


Magisus:
Never mind, segfault fixed!

MikaelSmith:
\o/

Magisus:
Alright, all classes are implemented and all tests pass. Ready for review!

lib/tests/config_node_test.cc
```diff
@@ -92,4 +94,240 @@ TEST_CASE("simple values", "[config-node]") {
...
+            line_node(10),
+            node_key_value_pair(node_key("baz.abc.ghi"), int_node(52)),
+            line_node(11),
+            node_key_value_pair(node_key("baz.abc.ghi"), int_node(53)),
+            line_node(12),
+            close_brace_node()
+    });
+    REQUIRE(orig_text == original_node->render());
+
+    string final_text = string("foo : bar\nbaz : {\n\t\"abc.def\" : true\n\t") +
```
MikaelSmith: Instead of concatenating strings, you can concat string literals, as in
```
"foo : bar\n" "baz : "
"{\n\t\"..."
```
This is C-style string literal concatenation, and can span multiple lines.

lib/tests/config_node_test.cc
```diff
@@ -92,4 +94,240 @@ TEST_CASE("simple values", "[config-node]") {
...
+            string("//this is a comment about the below setting") +
+            string("\n\n\tabc : {\n\t\tdef : false\n\t\t\n\t\t\"this.does.not.exist@@@+$#\"") +
+            string(" : {\n\t\t  end : doesnotexist\n\t\t}\n\t}\n}\n\nbaz.abc.ghi : randomunquotedString\n}");
```
MikaelSmith: That last `}` is weird, but definitely matches the original test.

MikaelSmith:
Any notes you could add to describe what each class is would be helpful. Not sure what a field is.

lib/src/nodes/abstract_config_node.cc
```diff
@@ -1,4 +1,6 @@
 #include <internal/nodes/abstract_config_node.hpp>
+#include <iostream>
```
MikaelSmith: Why were these needed?

lib/src/nodes/config_node_object.cc
```diff
@@ -0,0 +1,320 @@
...
+
+    bool config_node_object::has_value(path desired_path) const {
+        for (auto&& node : children()) {
+            if (auto field = dynamic_pointer_cast<config_node_field>(node)) {
+                path key = field->path()->get_path();
+                if (key == desired_path || key.starts_with(desired_path)) {
+                    return true;
+                } else if (desired_path.starts_with(key)) {
+                    if (auto object = dynamic_pointer_cast<config_node_object>(field->get_value())) {
+                        path remaning_path = desired_path.sub_path(key.length());
```
MikaelSmith: Typo in 'remaning_path'.

lib/src/nodes/config_node_object.cc
```diff
@@ -0,0 +1,320 @@
...
+                    }
+                }
+            }
+        }
+        return false;
+    }
+
+    // TODO: Just gonna throw this out here, this method is disgustingly long and hard to read...
+    // (line for line port from Java)
+    shared_ptr<config_node_object> config_node_object::change_value_on_path(path desired_path, shared_node_value value,
```
MikaelSmith: Candidate for refactoring if we have to make significant changes to it (bugs, new features). Probably not worth it until then, because we'd also want to change the other projects.

Magisus:
I only have a vague idea of what the classes represent myself. There is no description in the original code. But I'll give it a shot based on the tests.

lib/src/path.cc
```diff
@@ -112,13 +112,17 @@ namespace hocon {
             count--;
             builder.append_key(*from.first());
             from = from.remainder();
-            if (from.has_remainder()) {
+            if (from.empty()) {
```
MikaelSmith: Wouldn't this be the same as `from == path()`?

lib/src/path.cc
```diff
@@ -112,13 +112,17 @@ namespace hocon {
...
                 throw config_exception("sub_path last_index out of range");
             }
         }
         return builder.result();
     }
 
+    bool path::empty() const {
+        return first() == nullptr;
```
MikaelSmith: This seems like it could just return __path.isEmpty().

lib/src/path_parser.cc
```diff
@@ -143,14 +143,18 @@ namespace hocon {
             return token_list { t };
         }
 
+        // The split iterator tacks an empty string onto the end of the iterator if the string
+        // ends in the split character; so we need to remove trailing '.'
+        string text_to_split = boost::trim_right_copy_if(token_text, is_any_of("."));
```
MikaelSmith: Line 165 seems like it duplicates this. Why didn't it work? (also it could've used `token_text.back()`).

MikaelSmith: And if we don't need that code, we could just drop the const on token_text and trim it in-place.

lib/src/path_parser.cc
```diff
@@ -143,14 +143,18 @@ namespace hocon {
...
+
         auto token_it = boost::make_iterator_range(make_split_iterator(
-                token_text, token_finder(is_any_of("."))), split_iterator<string::const_iterator>());
+                text_to_split, token_finder(is_any_of("."))), split_iterator<string::iterator>());
 
         token_list split_tokens;
         for (auto& s : token_it) {
             string token_string { s.begin(), s.end() };
```
MikaelSmith: Is this really needed? I'd think s would implicitly convert to string as needed for the make_shared calls.

lib/src/path_parser.cc
Magisus: So these address two separate issues: line 165 handles the case where token_text is e.g. "a.b.c", in which c is considered a key to a final value, I believe (not 100% sure, but that's kind of what the tests look like). My problem was in the case of "a.b.c.", where the split iterator returned an extra empty string at the end, resulting in double "." when using the tokens to make a path.

lib/src/path_parser.cc
Magisus: Apparently it doesn't, I get a compile error.

lib/src/path_parser.cc
MikaelSmith: Ah, I misunderstood what it was checking (`!=`, not `==`). Also double checked Java's [string.split](http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#split(java.lang.String)) implementation to verify it leaves off trailing empty strings.

Magisus: I'll put in a second comment to that effect.

MikaelSmith: We could avoid copying the string by doing the `!= '.'` before-hand and saving it as a boolean.

MikaelSmith:
This looks reasonable. :+1:

PR #12, opened 2015-08-21
(HC-7) Port the ConfigDocumentParser

Magisus:
Adds the main parser class, used for the production of ConfigDocuments, which are representations of HOCON files in code.

### DIFF
--- a/lib/inc/internal/config_document_parser.hpp
+++ b/lib/inc/internal/config_document_parser.hpp
@@ -1,0 +1,84 @@
+#pragma once
+
+#include <hocon/config_parse_options.hpp>
+#include <internal/nodes/config_node_object.hpp>
+#include <internal/nodes/config_node_root.hpp>
+#include <internal/nodes/config_node_include.hpp>
+#include <internal/simple_config_origin.hpp>
+#include <internal/tokenizer.hpp>
+
+#include <stack>
+
+namespace hocon { namespace config_document_parser {
+
+    class parse_exception : public std::runtime_error {
+    public:
+        parse_exception(std::string const& message) : runtime_error(message) { }
+    };
+
+    std::shared_ptr<config_node_root> parse(token_iterator tokens,
+                                                   shared_origin origin, config_parse_options options);
+
+    shared_node_value parse_value(token_iterator tokens,
+                                         shared_origin origin, config_parse_options options);
+
+    class parse_context {
+    public:
+        parse_context(config_syntax flavor, shared_origin origin, token_iterator tokens);
+
+        std::shared_ptr<config_node_root> parse();
+
+        /**
+         * Parse a given input stream into a single value node. Used when doing a replace inside a ConfigDocument.
+         */
+        shared_node_value parse_single_value();
+
+    private:
+        shared_token pop_token();
+        shared_token next_token();
+        shared_token next_token_collecting_whitespace(shared_node_list& nodes);
+        void put_back(shared_token token);
+
+        /**
+         * In arrays and objects, comma can be omitted
+         * as long as there's at least one newline instead.
+         * this skips any newlines in front of a comma,
+         * skips the comma, and returns true if it found
+         * either a newline or a comma. The iterator
+         * is left just after the comma or the newline.
+         */
+        bool check_element_separator(shared_node_list& nodes);
+
+        /** Parse a concatenation. If there is no concatenation, return the next value. */
+        shared_node_value consolidate_values(shared_node_list& nodes);
+
+        std::string add_quote_suggestion(std::string bad_token, std::string message,
+                                         bool inside_equals, path* last_path);
+
+        std::string add_quote_suggestion(std::string bad_token, std::string message);
+
+        shared_node_value parse_value(shared_token t);
+        std::shared_ptr<config_node_path> parse_key(shared_token t);
+        bool is_key_value_separator(shared_token t);
+        std::shared_ptr<config_node_include> parse_include(shared_node_list& children);
+        std::shared_ptr<config_node_complex_value> parse_object(bool had_open_curly);
+        std::shared_ptr<config_node_complex_value> parse_array();
+
+        static bool is_include_keyword(shared_token t);
+        static bool is_unquoted_whitespace(shared_token t);
+        static bool is_valid_array_element(shared_token t);
+
+        int _line_number;
+        std::stack<shared_token> _buffer;
+        token_iterator _tokens;
+        config_syntax _flavor;
+        shared_origin _base_origin;
+
+        // this is the number of "equals" we are inside,
+        // used to modify the error message to reflect that
+        // someone may think this is .properties format.
+        int _equals_count;
+    };
+
+
+}}  // namespace hocon::config_document_parser
--- a/lib/src/config_document_parser.cc
+++ b/lib/src/config_document_parser.cc
@@ -1,0 +1,630 @@
+#include <internal/config_document_parser.hpp>
+#include <internal/nodes/config_node_single_token.hpp>
+#include <internal/nodes/config_node_comment.hpp>
+#include <internal/nodes/config_node_concatenation.hpp>
+#include <internal/nodes/config_node_simple_value.hpp>
+#include <internal/path_parser.hpp>
+#include <internal/config_util.hpp>
+
+#include <unordered_map>
+#include <internal/nodes/config_node_field.hpp>
+#include <internal/nodes/config_node_array.hpp>
+
+using namespace std;
+
+namespace hocon { namespace config_document_parser {
+
+    shared_ptr<config_node_root> parse(token_iterator tokens, shared_origin origin,
+                                                                    config_parse_options options)
+    {
+        parse_context context { options.get_syntax(), move(origin), move(tokens) };
+        return context.parse();
+    }
+
+    shared_node_value parse_value(token_iterator tokens, shared_origin origin,
+                                                          config_parse_options options)
+    {
+        parse_context context { options.get_syntax(), move(origin), move(tokens) };
+        return context.parse_single_value();
+    }
+
+    /** Parse context */
+    parse_context::parse_context(config_syntax flavor, shared_origin origin, token_iterator tokens) :
+        _line_number(1), _tokens(move(tokens)), _flavor(flavor), _base_origin(move(origin)) { }
+
+    shared_token parse_context::pop_token() {
+        if (_buffer.empty()) {
+            return _tokens.next();
+        }
+        shared_token top = _buffer.top();
+        _buffer.pop();
+        return top;
+    }
+
+    shared_token parse_context::next_token() {
+        shared_token t = pop_token();
+        if (_flavor == config_syntax::JSON) {
+            if (t->get_token_type() == token_type::UNQUOTED_TEXT && !is_unquoted_whitespace(t)) {
+                throw parse_exception("Token not allowed in valid JSON: '" + t->token_text() + "'");
+            } else if (t->get_token_type() == token_type::SUBSTITUTION) {
+                throw parse_exception("Substitutions (${} syntax) not allowed in JSON");
+            }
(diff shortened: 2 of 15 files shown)


lib/inc/internal/config_document_parser.hpp
```diff
@@ -0,0 +1,83 @@
+#pragma once
+
+#include <hocon/config_parse_options.hpp>
+#include <internal/nodes/config_node_object.hpp>
+#include <internal/nodes/config_node_root.hpp>
+#include <internal/nodes/config_node_include.hpp>
+#include <internal/simple_config_origin.hpp>
+#include <internal/tokenizer.hpp>
+#include "tokenizer.hpp"
```
MikaelSmith: Lines 8 and 9 seem to point to the same thing.

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+                if (t->get_token_type() == token_type::NEWLINE) {
+                    _line_number = t->line_number() + 1;
+                }
+            } else if (t->get_token_type() == token_type::COMMENT) {
+                nodes.push_back(make_shared<config_node_comment>(t));
+            } else {
+                if (t->line_number() >= 0) {
+                    _line_number = t->line_number();
+                }
+                return t;
```
MikaelSmith: How does next_token signal there are no more tokens?

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+            shared_token t = next_token_collecting_whitespace(nodes);
+            if (t->get_token_type() == token_type::COMMA) {
+                nodes.push_back(make_shared<config_node_single_token>(t));
+                return true;
+            } else {
+                put_back(t);
+                return false;
+            }
+        } else {
+            bool saw_separator_or_newline = false;
```
MikaelSmith: Funny naming when it will only be true for a newline.

lib/src/config_document_parser.cc
Magisus: I'm not sure that it does. Most of the no-more-tokens checking happens elsewhere; this class's loops tend to terminate on other things, and assume there will always be another token. If there isn't, usually it's because the stream of tokens is malformed, and an exception is thrown elsewhere.

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+            return nullptr;
+        }
+
+        // create only if we have value tokens
+        shared_node_list values;
+        int value_count = 0;
+
+        // ignore a newline up front
+        shared_token t = next_token_collecting_whitespace(nodes);
+        while (t) {
```
MikaelSmith: `while(t)` instead of `while(true)`? Did this come up as a bug, because it differs from https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/impl/ConfigDocumentParser.java#L141?

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+        return v;
+    }
+
+    shared_ptr<config_node_path> parse_context::parse_key(shared_token token) {
+        if (_flavor == config_syntax::JSON) {
+            if (tokens::is_value_with_type(token, config_value_type::STRING)) {
+                single_token_iterator it(token);
+                return make_shared<config_node_path>(path_parser::parse_path_node_expression(it, nullptr));
+            } else {
+                throw parse_exception("\"Expecting close brace } or a field name here, got " + token->to_string());
```
MikaelSmith: The `\"` looks new compared to the original. https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/impl/ConfigDocumentParser.java#L266

lib/src/config_document_parser.cc
Magisus: Yeah it was a bug, caused because of some differences of implementation. Their type checking doesn't care about null (it's mostly instanceof on the backend), and/or the java library iterators behave differently. Before this change, this loop went infinite when using a single_token_iterator implementation that didn't return null, and crashed when using one that did. Since the other iterators should never return null before being manually stopped, this is essentially a while(true) loop for every other kind. See https://github.com/puppetlabs/cpp-hocon/pull/12/files#diff-61fed8ba62efd96b50205b88f8b122b6R23 for my dummy iterator implementations (these are built in to the collections used in the Java, e.g. tokenize() returns an Iterator<Token>).

lib/src/config_document_parser.cc
Magisus: Oh, CLion copy/paste fail.

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+
+        if (had_open_curly) {
+            object_nodes.push_back(make_shared<config_node_single_token>(tokens::open_curly_token()));
+        }
+
+        while (true) {
+            shared_token t = next_token_collecting_whitespace(object_nodes);
+            if (t->get_token_type() == token_type::CLOSE_CURLY) {
+                if (_flavor == config_syntax::JSON && after_comma) {
+                    throw parse_exception(add_quote_suggestion(t->to_string(),
```
MikaelSmith: This signature for add_quote_suggestion happens often enough it'd be nice to still have the shorter call signature.

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+                    if (!is_key_value_separator(after_key)) {
+                        throw parse_exception(add_quote_suggestion(after_key->to_string(),
+                        "Key '" + key_path->render() + "' may not be followed by token: " + after_key->to_string(),
+                        _equals_count > 0));
+                    }
+
+                    key_value_nodes.push_back(make_shared<config_node_single_token>(after_key));
+                    if (after_key->get_token_type() == token_type::EQUALS) {
+                        inside_equals = true;
+                        _equals_count++;
```
MikaelSmith: I'm not sure if the compiler optimizes this when the return value is ignored, but semantically doing `++__equals_count` is less work because it increments then returns that value. `_equals_count++` potentially has to save a temporary, increment the original, and return the temporary. It's a good habit to prefer the prefix notation.

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+                    _equals_count--;
+                }
+                last_inside_equals = inside_equals;
+
+                string key = *key_path->get_path().first();
+                path remaining = key_path->get_path().remainder();
+
+                if (remaining.empty()) {
+                    try {
+                        keys.at(key);
```
MikaelSmith: This is an expensive way to check for existence. There are two simpler ways, depending on what you need out of it
- Check for keys.find(key) == keys.end()
- Just insert/emplace, and check the return value. If the key already exists, the insert/emplace will not overwrite it.
http://www.cplusplus.com/reference/unordered_map/unordered_map/emplace/

So I'd rewrite this as
```
auto emplace = keys.emplace(move(key), true);
if (!emplace.second) {
    if (_flavor...) { ... }
    // This is a little unclear, but the first item is a pair<key,value> reference, so we can update the value by overwriting second.
    emplace.first.second = true;
}
```

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+            // just after a value
+            if (check_element_separator(children)) {
+                // comma or newline equivalent consumed
+            } else {
+                t = next_token_collecting_whitespace(children);
+                if (t->get_token_type() == token_type::CLOSE_SQUARE) {
+                    children.push_back(make_shared<config_node_single_token>(t));
+                    return make_shared<config_node_array>(children);
+                } else {
+                    throw parse_exception("List should have ended with ']' or had a comma, instead had token " +
```
MikaelSmith: Missing a `:`, https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/impl/ConfigDocumentParser.java#L533.

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+
+            // special case the first element
+            if (t->get_token_type() == token_type::CLOSE_SQUARE) {
+                children.push_back(make_shared<config_node_single_token>(t));
+                return make_shared<config_node_array>(children);
+            } else if (is_valid_array_element(t)) {
+                next_value = parse_value(t);
+                children.push_back(next_value);
+            } else {
+                throw parse_exception("List should have ']' or a first element after the '[', instead had token " +
```
MikaelSmith: Missing a `:`, https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/impl/ConfigDocumentParser.java#L514.

MikaelSmith: I'd keep the same error messages without a reason to change them.

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+                t->to_string() + " (if you want " + t->to_string() + " to be part of a string value, then"
+                "double quote it");
```
MikaelSmith: Dropped a `)` at the end of the string.

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+                // comma or newline equivalent consumed
+            } else {
+                t = next_token_collecting_whitespace(children);
+                if (t->get_token_type() == token_type::CLOSE_SQUARE) {
+                    children.push_back(make_shared<config_node_single_token>(t));
+                    return make_shared<config_node_array>(children);
+                } else {
+                    throw parse_exception("List should have ended with ']' or had a comma, instead had token " +
+                                          t->to_string() + " (if you want " + t->to_string() + " to be part of a "
+                                          "string value, then double quote it");
```
MikaelSmith: Dropped a `)` at the end of the string.

lib/src/config_document_parser.cc
```diff
@@ -0,0 +1,632 @@
...
+            } else {
+                t = next_token_collecting_whitespace(children);
+                if (is_valid_array_element(t)) {
+                    next_value = parse_value(t);
+                    children.push_back(next_value);
+                } else if (_flavor != config_syntax::JSON && t->get_token_type() == token_type::CLOSE_SQUARE) {
+                    // we allow one trailing comma
+                    put_back(t);
+                } else {
+                    throw parse_exception("List should have had a new element after a comma, instead had token " +
```
MikaelSmith: `:` and `)`.

lib/src/nodes/config_node_include.cc
```diff
@@ -5,19 +5,20 @@ using namespace std;
...
 
-    config_node_include::config_node_include(vector<unique_ptr<abstract_config_node>> children,
+    config_node_include::config_node_include(shared_node_list children,
                                              config_include_kind kind) :
         _children(move(children)), _kind(kind) { }
 
     token_list config_node_include::get_tokens() const {
         token_list tokens;
         for (auto&& node : _children) {
-            tokens.insert(tokens.end(), node->get_tokens().begin(), node->get_tokens().end());
```
MikaelSmith: :+1:

Magisus: Took me a long time to figure out how to fix the segfault this used to cause... XD

lib/src/config_document_parser.cc
MikaelSmith: This `:` is still missing.

MikaelSmith:
Otherwise :+1:

Magisus:
Done, commits squashed.

Magisus:
Also, do you know why Coveralls can't find the source files to show line-by-line coverage? I seem to remember this working in Leatherman...

MikaelSmith:
I think it's fixed now.

Magisus:
Woo! Just been curious if we're missing anything important, because while there a lot of tests, a lot of them feel like they just exercise the same pieces of code in slightly different orders....

PR #17, opened 2015-08-27
(maint) Use myget mirror for Windows packages

MikaelSmith:
Getting directly from cmake.org has been unreliable recently. Use the
myget mirrored packages.


MikaelSmith:
@Magisus can you merge this later? AppVeyor can't get from cmake.org for some reason, so need to use our mirror on all C++ projects.

PR #18, opened 2015-08-28
(HC-27) Port Config and SimpleConfig

Magisus:
Basic functionality for these classes. No tests written. They depend on the parser and merge functionality.

Some pretty major restructuring was needed in order to resolve some inheritance issues. Existing tests all pass.


MikaelSmith:
functional_list.hpp will need to move to inc/hocon as well. Later I may rework both path.hpp and functional_list.hpp so we don't need those public.

MikaelSmith:
Waiting on Travis.

MikaelSmith:
One last thought on this: C++ supports return type covarience, but shared_ptr messes it up. Some workarounds were proposed at http://stackoverflow.com/a/6925201, but I still prefer what we did - removed the need for it.

PR #22, opened 2015-12-28
(HC-40) Port parseable_not_found and relative_name_source classes

whopper:
These classes are needed by the simple_includer class. Bring in
ParseableNotFound, the NameSource interface and RelativeNameSource.


MikaelSmith:
Looks legit.

PR #23, opened 2015-12-29
(HC-10) Port Unmergeable interface and SubstitutionExpression class

whopper:
The first of these commits ports the small Unmergeable interface, which is needed by several other classes down the road.

The second commit ports the SubstitutionExpression class, which is needed by the ConfigReference implementation.

I'm putting this up now rather than waiting until all of ConfigReference is done both to make review more digestible and get quicker feedback on any insanity in this code before I get too deep.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -60,6 +60,7 @@ set(PROJECT_SOURCES
     src/config_resolve_options.cc
     src/config.cc
     src/default_transformer.cc
+    src/substitution_expression.cc
 )
 
 ## An object target is generated that can be used by both the library and test executable targets.
--- a/lib/inc/internal/substitution_expression.hpp
+++ b/lib/inc/internal/substitution_expression.hpp
@@ -1,0 +1,25 @@
+#pragma once
+
+#include <string>
+#include <memory>
+
+namespace hocon {
+
+    class substitution_expression : public std::enable_shared_from_this<substitution_expression> {
+    public:
+        substitution_expression(std::string path, bool optional);
+
+        std::string path() const;
+        bool optional() const;
+
+        std::shared_ptr<substitution_expression> change_path(std::string new_path);
+
+        std::string to_string() const;
+
+    private:
+        const std::string _path;
+        const bool _optional;
+    };
+
+}  // namespace hocon
+
--- a/lib/inc/internal/unmergeable.hpp
+++ b/lib/inc/internal/unmergeable.hpp
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <hocon/config_value.hpp>
+#include <vector>
+
+namespace hocon {
+
+    /**
+    * Interface that tags a ConfigValue that is not mergeable until after
+    * substitutions are resolved. Basically these are special ConfigValue that
+    * never appear in a resolved tree, like {@link ConfigSubstitution} and
+    * {@link ConfigDelayedMerge}.
+    */
+    class unmergeable {
+    public:
+        virtual std::vector<shared_value> unmerged_values() const = 0;
+    };
+
+}  // namespace hocon
+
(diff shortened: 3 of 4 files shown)


lib/inc/internal/substitution_expression.hpp
```diff
@@ -0,0 +1,34 @@
...
+        substitution_expression(std::string path, bool optional);
+
+        std::string path() const;
+        bool optional() const;
+
+        std::shared_ptr<substitution_expression> change_path(std::string new_path);
+
+        std::string to_string() const;
+
+        template <typename T>
```
whopper: This is to mimic the Java implementation which takes any type of `Object` to compare. I.e,
```
    @Override
    public boolean equals(Object other) {
        if (other instanceof SubstitutionExpression) {
...
```

lib/inc/internal/unmergeable.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+    /**
+    * Interface that tags a ConfigValue that is not mergeable until after
+    * substitutions are resolved. Basically these are special ConfigValue that
+    * never appear in a resolved tree, like {@link ConfigSubstitution} and
+    * {@link ConfigDelayedMerge}.
+    */
+    class unmergeable {
+    public:
+        virtual std::vector<shared_value> unmerged_values() = 0;
```
whopper: The Java implementation does this:
```
Collection<? extends AbstractConfigValue> unmergedValues();
```

I *think* this means a collection of the type of any of `AbstractConfigValue`'s children, so is a vector of pointers to the base `config_value`s right here?

whopper:
Note: https://github.com/puppetlabs/cpp-hocon/pull/24 needs to be merged first and this rebased.

lib/inc/internal/substitution_expression.hpp
MikaelSmith: Java uses equals because `==` means exact object equality, and can't be overridden. In C++ if you want to check two objects are the same object you compare their pointers. We also don't need to support general comparison, we shouldn't be trying to compare a `substitution_expression` to a different class.

cpp-hocon has used `operator==` overloading, as in https://github.com/puppetlabs/cpp-hocon/blob/master/lib/inc/hocon/config.hpp#L588 (for limited access) or https://github.com/puppetlabs/cpp-hocon/blob/master/lib/inc/hocon/parser/config_document.hpp#L92 for global access.

lib/src/substitution_expression.cc
```diff
@@ -0,0 +1,32 @@
...
+
+    bool substitution_expression::optional() const {
+        return _optional;
+    }
+
+    shared_ptr<substitution_expression> substitution_expression::change_path(std::string new_path) {
+        if (new_path == _path) {
+            return shared_from_this();
+        } else {
+            return make_shared<substitution_expression>(move(new_path), _optional);
```
MikaelSmith: :+1:

lib/src/substitution_expression.cc
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+
+    string substitution_expression::to_string() const {
+        string result;
+        return result.append(string("${") + (_optional ? "?" : "") + _path + "}");
```
MikaelSmith: I'd probably skip `append` here, just do `return string("${") + (_optional ? "?" : "") + _path + "}";`.

whopper:
@MikaelSmith awesome, thanks for the review. Updated based on your comments!

lib/CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ set(PROJECT_SOURCES
     src/config_resolve_options.cc
     src/config.cc
     src/default_transformer.cc
+    src/substitution_expression
```
MikaelSmith: Does this work? You might need to add `.cc` to the end.

whopper: Hah oh weird, I think CLion automatically added it like that - updated.

PR #24, opened 2015-12-29
(maint) Fixup issues with partial ConfigParser implementation

whopper:
This commit changes the definition of simple_config_origin::append_comments
to match its declaration. In addition, several erroneous #include paths
have been fixed in the config_parser header.


whopper:
Note: merge this before any of my other PRs.

PR #25, opened 2015-12-30
(HC-41) Port Merge-related classes and ConfigValue::merge* methods

whopper:
The first of these commits ports skeletons of the configDelayedMergeObject
and configDelayedMerge classes, as well as the mergeableValue
interface. These are all needed for the work in the second commit, which ports each of the config  merge-related methods of ConfigValue, which are needed by the ConfigConcatenation class. They include:

```
ConfigValue::mergedWithTheUnmergeable,
ConfigValue::MergedWithObject,
ConfigValue::MergedWithNonObject,
ConfigValue::ConstructDelayedMerge,
ConfigValue::withFallback
```

Also ported are various other things necessary for ConfigConcatenation, including:

```
SimpleConfigOrigin::mergeOrigins
SimpleConfigOrigin::mergeTwo
ConfigObject::mergeOrigins
ConfigImplUtil class
```


whopper:
Note: https://github.com/puppetlabs/cpp-hocon/pull/24 needs to be merged first and this rebased.

lib/inc/internal/config_impl_util.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <string>
+#include <vector>
+
+namespace hocon {
+
+    class config_impl_util {
+    public:
+        static bool equals_handling_null(std::string const& a, std::string const& b);
```
whopper: Question here: the original Java implementation of these methods is generic:

```
    static boolean equalsHandlingNull(Object a, Object b) {
        if (a == null && b != null)
            return false;
        else if (a != null && b == null)
            return false;
        else if (a == b) // catches null == null plus optimizes identity case
            return true;
        else
            return a.equals(b);
    }
```

From what I've seen so far, it looks like we're only going to do this with strings and vectors, so I opted to just overload it for those types. Is this okay, or do we need to take a more generic approach?

MikaelSmith: Since this is only used for strings and vectors, I'm not sure it's needed. They'll never be null because we're using references, so we can just do `a == b` and call it a day.

lib/inc/hocon/config_value.hpp
```diff
@@ -133,6 +136,8 @@ namespace hocon {
         virtual resolve_status get_resolve_status() const;
 
         friend resolve_status resolve_status_from_values(std::vector<shared_value> const& v);
+        // this is only overridden to change the return type
+        shared_value with_fallback(std::shared_ptr<const config_mergeable> mergeable);
```
MikaelSmith: config_value doesn't inherit from anything, so what's it overriding?

lib/inc/hocon/config_object.hpp
```diff
@@ -27,6 +28,9 @@ namespace hocon {
         virtual shared_object with_value(path raw_path, shared_value value) const = 0;
         virtual shared_object with_value(std::string key, shared_value value) const = 0;
 
+        shared_value with_fallback(std::shared_ptr<const config_mergeable> mergeable);
+        static shared_origin merge_origins(std::vector<shared_value> stack);
```
MikaelSmith: mergeOrigins is part of the internal implementation, so it should go in the `protected` block.

lib/inc/hocon/config_mergeable.hpp
```diff
@@ -6,6 +6,7 @@
 namespace hocon {
 
     class LIBCPP_HOCON_EXPORT config_mergeable {
+        friend class config_value;
```
MikaelSmith: Why was this needed?

whopper: Because of one place in config_value where we do:

```
auto other = dynamic_pointer_cast<const config_mergeable>(mergeable)->to_fallback_value();
```

and if config_value isn't a friend class, we get:

```
/Users/whopper/Coding/cpp-hocon/lib/src/values/config_value.cc:124:83: error: 'to_fallback_value' is a protected member of 'hocon::config_mergeable'
            auto other = dynamic_pointer_cast<const config_mergeable>(mergeable)->to_fallback_value();
                                                                                  ^
/Users/whopper/Coding/cpp-hocon/lib/inc/hocon/config_mergeable.hpp:68:30: note: can only access this member on an object of type 'hocon::config_value'
        virtual shared_value to_fallback_value() const = 0;
```

whopper: @MikaelSmith that cast and/or statement may also just be wrong :P

MikaelSmith: Ok, that maybe makes sense. I'm going to play with this a bit, we shouldn't need to `friend class` when calling a method of a class the method inherits from.

MikaelSmith: http://stackoverflow.com/a/4672888 answers the friend class requirement. We can revisit later if we'd like to just make some methods public instead of using friend everywhere.

lib/inc/internal/config_delayed_merge.hpp
```diff
@@ -0,0 +1,21 @@
...
+#pragma once
+
+namespace hocon {
+
+    class config_delayed_merge : public config_value, public unmergeable {
+    public:
+        config_delayed_merge(shared_origin origin, std::vector<shared_value> stack);
+
+        config_value_type value_type() const override;
+        std::vector<shared_value> unmerged_values() const override;
```
MikaelSmith: I wonder if this (and thus the unmergeable interface) should return a `const&`. Not clear yet.

lib/inc/internal/simple_config_origin.hpp
```diff
@@ -31,10 +32,28 @@ namespace hocon {
 
         std::shared_ptr<const simple_config_origin> append_comments(std::vector<std::string> comments) const;
 
+        static shared_origin merge_origins(shared_origin a, shared_origin b);
+        static shared_origin merge_origins(std::vector<shared_value> stack);
```
MikaelSmith: These vectors should be passed by `const&`.

whopper: Updated.

lib/inc/internal/simple_config_origin.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <hocon/config_origin.hpp>
+#include <hocon/config_value.hpp>
```
MikaelSmith: Why is this header file needed? It wasn't before, and nothing you added should require it.

lib/src/config_delayed_merge.cc
```diff
@@ -0,0 +1,31 @@
+#include <internal/config_delayed_merge.hpp>
+#include <internal/config_delayed_merge_object.hpp>
+#include <internal/config_exception.hpp>
+
+using namespace std;
+
+namespace hocon {
+
+    config_delayed_merge::config_delayed_merge(shared_origin origin, std::vector<shared_value> stack) :
+        config_value(move(origin)), _stack(stack) {
```
MikaelSmith: Should `move(stack)`.

lib/src/config_delayed_merge.cc
```diff
@@ -0,0 +1,31 @@
...
+                throw config_exception("placed nested delayed_merge in a config_delayed_merge, should have consolidated stack");
+            }
+        }
+    }
+
+    config_value_type config_delayed_merge::value_type() const {
+        throw config_exception("called value_type() on value with unresolved substitutions, need to config#resolve() first, see API docs");
+    }
+
+    vector<shared_value> config_delayed_merge::unmerged_values() const {
```
MikaelSmith: Yeah, this seems like it should return a `const &` to the vector, not a copy of the vector.

lib/src/objects/config_object.cc
```diff
@@ -51,4 +52,42 @@ namespace hocon {
         return config_value_type::OBJECT;
     }
 
+    shared_ptr<const config_mergeable> config_object::with_fallback(shared_ptr<const config_mergeable> mergeable) const {
+        return dynamic_pointer_cast<const config_value>(config_value::with_fallback(mergeable));
```
MikaelSmith: The dynamic_pointer_cast shouldn't be necessary. So we should be able to get rid of the override and fall back to the config_value version.

lib/src/objects/config_object.cc
```diff
@@ -51,4 +52,42 @@ namespace hocon {
...
+            }
+
+            auto cv = dynamic_pointer_cast<const config_object>(v);
+            if (cv && cv->get_resolve_status() == resolve_status::RESOLVED && cv->is_empty()) {
+                // don't include empty files or the .empty()
+                // config in the description, since they are
+                // likely to be "implementation details"
+            } else {
+                origins.push_back(v->origin());
+                num_merged += 1;
```
MikaelSmith: I know it mirrors the original, but origins.size() will always be equal to num_merged. We could do away with tracking it separately and check `origins.empty()` at line 84.

lib/src/simple_config_origin.cc
```diff
@@ -54,6 +57,171 @@ namespace hocon {
         }
     }
 
+    shared_origin simple_config_origin::merge_origins(shared_origin a, shared_origin b) {
+        return merge_two(dynamic_pointer_cast<const simple_config_origin>(a), dynamic_pointer_cast<const simple_config_origin>(b));
+    }
+
+    shared_origin simple_config_origin::merge_origins(std::vector<shared_value> const& stack) {
+        vector<shared_origin> origins;
```
MikaelSmith: It's probably worth doing `origins.reserve(stack.size())`, to ensure we have the right amount of memory and don't have to realloc later.

lib/src/simple_config_origin.cc
```diff
@@ -54,6 +57,171 @@ namespace hocon {
...
+        }
+
+        return merge_origins(origins);
+    }
+
+    shared_origin simple_config_origin::merge_origins(std::vector<shared_origin> const& stack) {
+        if (stack.empty()) {
+            throw config_exception("can't merge empty list of origins");
+        } else if (stack.size() == 1) {
+            return stack.front();
```
MikaelSmith: Is this equivalent to `stack.iterator().next()`?

lib/src/simple_config_origin.cc
```diff
@@ -54,6 +57,171 @@ namespace hocon {
...
+        int merged_end_line;
+
+        origin_type merged_type;
+        if (a->_origin_type == b->_origin_type) {
+            merged_type = a->_origin_type;
+        } else {
+            merged_type = origin_type::GENERIC;
+        }
+
+        const string MERGE_OF_PREFIX { "merge of " };
```
MikaelSmith: Might as well make it `static`, to match the original.

lib/src/simple_config_origin.cc
```diff
@@ -54,6 +57,171 @@ namespace hocon {
...
+
+        if (a_desc == b_desc) {
+            merged_desc = a_desc;
+
+            if (a->line_number() < 0) {
+                merged_start_line = b->line_number();
+            } else if (b->line_number() < 0) {
+                merged_start_line = a->line_number();
+            } else {
+                merged_start_line = max(a->_end_line_number, b->_end_line_number);
```
MikaelSmith: This should be `min`, not `max`.

lib/src/simple_config_origin.cc
```diff
@@ -54,6 +57,171 @@ namespace hocon {
...
+
+            merged_end_line = max(a->_end_line_number, b->_end_line_number);
+        } else {
+            // this whole merge song-and-dance was intended to avoid this case
+            // whenever possible, but we've lost. Now we have to lose some
+            // structured information and cram into a string.
+
+            // description() method includes line numbers, so use it instead
+            // of description field.
+            string a_full = a->description();
```
MikaelSmith: This was already done for a_desc and b_desc above, I think we can just use them.

lib/src/simple_config_origin.cc
```diff
@@ -54,6 +57,171 @@ namespace hocon {
...
+            merged_desc = MERGE_OF_PREFIX + a_full + "," + b_full;
+            merged_start_line = -1;
+            merged_end_line = -1;
+        }
+
+        string merged_resource;
+        if (a->_resource_or_null == b->_resource_or_null) {
+            merged_resource = a->_resource_or_null;
+        } else {
+            merged_resource = "";
```
MikaelSmith: Can remove the else clause, that's the default initialized behavior for merged_resource.

lib/inc/hocon/config_value.hpp
```diff
@@ -163,8 +169,32 @@ namespace hocon {
...
+                                                 std::shared_ptr<const unmergeable> fallback) const;
+        shared_value merged_with_the_unmergeable(std::shared_ptr<const unmergeable> fallback) const;
+
+        shared_value merged_with_object(std::vector<shared_value> const& stack, shared_object fallback) const;
+        shared_value merged_with_object(shared_object fallback) const;
+
+        shared_value merged_with_non_object(std::vector<shared_value> const& stack, shared_value fallback) const;
+        shared_value merged_with_non_object(shared_value fallback) const;
+
+        shared_value construct_delayed_merge(shared_origin origin, std::vector<shared_value> stack) const;
```
MikaelSmith: The only reason this function exists is to be overridden by config_object. So it should be virtual.

lib/src/values/config_value.cc
```diff
@@ -152,4 +154,112 @@ namespace hocon {
...
+                                                           std::shared_ptr<const unmergeable> fallback) const {
+        require_not_ignoring_fallbacks();
+
+        // if we turn out to be an object, and the fallback also does,
+        // then a merge may be required; delay until we resolve.
+        std::vector<shared_value> new_stack { stack };
+
+        auto unmerged_values = fallback->unmerged_values();
+        new_stack.insert(new_stack.end(), unmerged_values.begin(), unmerged_values.end());
+        return construct_delayed_merge(config_object::merge_origins(new_stack), new_stack);
```
MikaelSmith: Use `move(new_stack)`.

lib/src/values/config_value.cc
```diff
@@ -152,4 +154,112 @@ namespace hocon {
...
+
+        if (get_resolve_status() == resolve_status::RESOLVED) {
+            // falling back to a non-object doesn't merge anything, and also
+            // prohibits merging any objects that we fall back to later.
+            // so we have to switch to ignoresFallbacks mode.
+            return with_fallbacks_ignored();
+        } else {
+            // if unresolved, we may have to look back to fallbacks as part of
+            // the resolution process, so always delay
+            return delay_merge(move(stack), move(fallback));
```
MikaelSmith: You don't want to `move` here, it's passing through a reference.

lib/inc/internal/simple_config_origin.hpp
whopper: Ah, looks like I added it because of one of the implementations of `simple_config_origin::merge_origins`, with:

```
    shared_origin simple_config_origin::merge_origins(std::vector<shared_value> const& stack) {
        vector<shared_origin> origins;
        for (auto& v : stack) {
            origins.push_back(v->origin());
        }
```
If I didn't include it, I'd get 'error: member access into incomplete type 'element_type' (aka 'const hocon::config_value') origins.push_back(v->origin());`

lib/src/values/config_value.cc
```diff
@@ -152,4 +154,112 @@ namespace hocon {
...
+    }
+
+    shared_value config_value::merged_with_object(vector<shared_value> const& stack, shared_object fallback) const {
+        require_not_ignoring_fallbacks();
+
+        if (dynamic_pointer_cast<const config_object>(shared_from_this())) {
+            throw config_exception("Objects must reimplement merged_with_object");
+        }
+
+        return merged_with_non_object(move(stack), move(fallback));
```
MikaelSmith: Don't move, passing a reference.

lib/src/values/config_value.cc
```diff
@@ -5,6 +5,8 @@
 #include <internal/objects/simple_config_object.hpp>
 #include <internal/simple_config_origin.hpp>
 #include <internal/container.hpp>
+#include <internal/config_delayed_merge.hpp>
+#include <internal/mergeable_value.hpp>
```
MikaelSmith: This header was removed, need to remove the `#include` (Travis is failing).

lib/inc/internal/simple_config_origin.hpp
MikaelSmith: Yes. Keep header files included in other header files to a minimum, otherwise you really kill compile time.

lib/src/values/config_value.cc
whopper: Whoops, fixed!

lib/src/config_delayed_merge_object.cc
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    bool config_delayed_merge_object::is_empty() const {
+        // TODO
+        return {};
+    }
+
+    unordered_map<string, shared_value> const& config_delayed_merge_object::entry_set() const {
+        // TODO
+        return move(unordered_map<string, shared_value> {});
```
MikaelSmith: You don't need move here, that's implicit for an object created in the return call.

MikaelSmith:
Ok, I think that's good. :+1:

PR #26, opened 2016-01-05
(maint) Fix typo in README

HAIL9000:
Change `make tests` to `make test`.


whopper:
Aw yeah

There is one other thing that Facter specifies too:

```
For verbose test output, run ctest instead of using the test target:

$ cd release
$ ctest -V
```

Maybe we should add that too?

PR #27, opened 2016-01-07
(maint) Refactor config_origin, config_include_context, and config_document

MikaelSmith:
Replace use of shared_ptr<const config_origin> everywhere to provide
better encapsulation. Instead, use a config_origin class, where the
implementation is private (for now using a shared_ptr to impl to share
object data between identical origins, which is safe because they're
immutable). Use of the private API will be restricted via symbol
visibility.


MikaelSmith:
I may change this to make the INTERNAL API sections protected, and use friend class declarations. That appears to be what @Magisus started with config_object and config_value.

branan:
@MikaelSmith This is conflicting everywhere at this point

MikaelSmith:
Pretty much. Going to close it for now, maybe we'll decide to pick it up again sometime.

PR #28, opened 2016-01-07
[WIP] (HC-39) Port header files for SimpleConfigList

HAIL9000:
In order to port SimpleConfig List, add all the header files that
are needed. This includes the header file for SimpleConfigList and
several other classes that it relies on. At the moment, the header
files for resolve_context, resolve_result, and resolve_source are
just stubs without any methods or instance variables.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -61,6 +61,7 @@ set(PROJECT_SOURCES
     src/config.cc
     src/default_transformer.cc
     src/substitution_expression.cc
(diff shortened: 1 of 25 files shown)


MikaelSmith:
Closing in favor of https://github.com/puppetlabs/cpp-hocon/pull/30, so we can keep working on things that rely on this work.

Removed the unsupported_operation_exception, as that's part of Java core and only used to disable functions defined to satisfy the java.lang.List interface. We don't need to satisfy that interface, so I deleted the functions.

PR #30, opened 2016-01-08
(HC-39) Port header files for SimpleConfigList

MikaelSmith:
In order to port SimpleConfig List, add all the header files that
are needed. This includes the header file for SimpleConfigList and
several other classes that it relies on. At the moment, the header
files for resolve_context, resolve_result, and resolve_source are
just stubs without any methods or instance variables.

Also implements stubs for methods so everything links.


MikaelSmith:
ping @whopper

PR #31, opened 2016-01-09
(HC-39) Implement part of simple_config_list

MikaelSmith:


lib/inc/internal/simple_config_list.hpp
```diff
@@ -13,11 +13,11 @@
 
 namespace hocon {
 
-    class simple_config_list : public config_list, public container, public std::enable_shared_from_this<simple_config_list> {
```
MikaelSmith: Changed because it's considered a bad pattern to inherit from it multiple times. For some reason it started causing compile issues when I started using resolve_result.

PR #32, opened 2016-01-12
(maint) Ensure all files in CMakeLists.txt include file extension

whopper:


MikaelSmith:
Oh right, cmake.

kylog:
Was there a conscious reason that file didn't have an extension? Just curious.

whopper:
@kylog that file just got added in in the big PR Michael merged, and I *think* clion automatically adds files to the list without the extension. That or I just forgot to type it somehow :)

PR #33, opened 2016-01-12
(HC-39) Implement part of resolve_substitutions

MikaelSmith:
Implements portions of resolve_substitutions in existing classes
(config_value, simple_config_list, and simple_config_object); with
stubs of methods needed from resolve_context and resolve_source.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -64,6 +64,8 @@ set(PROJECT_SOURCES
     src/simple_config_list.cc
     src/config_delayed_merge.cc
     src/config_delayed_merge_object.cc
+    src/resolve_context.cc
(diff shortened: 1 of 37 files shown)


MikaelSmith:
ping @whopper

MikaelSmith:
One of my guiding principle is we should avoid pointers unless we're dealing with class hierarchies (config_value). If we want to share some implementation, pass objects by value and for expensive copies use shared_ptrs for the private members. So I'm passing resolve_source and resolve_context by value.

MikaelSmith:
Almost done, this could be merged as-is. There are a few helper methods that still need to be implemented.

whopper:
Woo :+1:

PR #34, opened 2016-01-14
(HC-35) Port complete config_concatenation class

whopper:
The first of these two commits also adds a few needed members and methods to the resolve_context and resolve_result classes.

### DIFF
--- a/lib/inc/internal/config_concatenation.hpp
+++ b/lib/inc/internal/config_concatenation.hpp
@@ -1,0 +1,57 @@
+#pragma once
+
+#include <hocon/config_value.hpp>
+#include <internal/unmergeable.hpp>
+#include <internal/container.hpp>
+#include <string>
+#include <vector>
+
+namespace hocon {
+
+    class config_exception;
+
+    /**
+     * A ConfigConcatenation represents a list of values to be concatenated (see the
+     * spec). It only has to exist if at least one value is an unresolved
+     * substitution, otherwise we could go ahead and collapse the list into a single
+     * value.
+     *
+     * Right now this is always a list of strings and ${} references, but in the
+     * future should support a list of ConfigList. We may also support
+     * concatenations of objects, but ConfigDelayedMerge should be used for that
+     * since a concat of objects really will merge, not concatenate.
+     */
+    class config_concatenation : public config_value, public unmergeable, public container {
+    public:
+        config_concatenation(shared_origin origin, std::vector<shared_value> pieces);
+
+        config_value_type value_type() const override;
+        std::vector<shared_value> unmerged_values() const override;
+
+        resolve_status get_resolve_status() const override;
+
+        shared_value replace_child(shared_value const& child, shared_value replacement) const override;
+        bool has_descendant(shared_value const& descendant) const override;
+        resolve_result<shared_value> resolve_substitutions(resolve_context const& context, resolve_source const& source) const override;
+
+        static std::vector<shared_value> consolidate(std::vector<shared_value> pieces);
+        static shared_value concatenate(std::vector<shared_value> pieces);
+        shared_value relativized(std::string prefix) const;
+
+        bool operator==(config_concatenation const& other) const;
+
+    protected:
+        shared_value new_copy(shared_origin origin) const override;
+        bool ignores_fallbacks() const;
+        void render(std::string& result, int indent, bool at_root, config_render_options options) const override;
+
+    private:
+        std::vector<shared_value> _pieces;
+
+        config_exception not_resolved() const;
+        static bool is_ignored_whitespace(shared_value value);
+        static void join(std::vector<shared_value> & builder, shared_value right);
+    };
+
+}  // namespace hocon
+
--- a/lib/src/config_concatenation.cc
+++ b/lib/src/config_concatenation.cc
@@ -1,0 +1,246 @@
+#include <hocon/config_object.hpp>
+#include <hocon/path.hpp>
+#include <internal/config_concatenation.hpp>
+#include <internal/values/config_string.hpp>
+#include <internal/default_transformer.hpp>
+#include <internal/simple_config_list.hpp>
+#include <internal/simple_config_origin.hpp>
+#include <internal/config_exception.hpp>
+#include <internal/values/config_string.hpp>
+#include <internal/resolve_result.hpp>
+#include <internal/resolve_source.hpp>
+#include <internal/resolve_context.hpp>
+
+using namespace std;
+
+namespace hocon {
+
+    config_concatenation::config_concatenation(shared_origin origin, std::vector<shared_value> pieces) :
+            config_value(move(origin)), _pieces(move(pieces)) {
+        if (_pieces.size() < 2) {
+            throw config_exception("Created concatenation with less than 2 items");
+        }
+
+        bool had_unmergeable = false;
+        for (shared_value p : _pieces) {
+            if (dynamic_pointer_cast<const config_concatenation>(p)) {
+                throw config_exception("config_concatenation should never be nested");
+            }
+
+            if (dynamic_pointer_cast<const unmergeable>(p)) {
+                had_unmergeable = true;
+            }
+        }
+
+        if (!had_unmergeable) {
+            throw config_exception("Created concatenation without an unmergeable in it");
+        }
+    }
+
+    config_value_type config_concatenation::value_type() const {
+        throw not_resolved();
+    }
+
+    vector<shared_value> config_concatenation::unmerged_values() const {
+        return { shared_from_this() };
+    }
+
+    resolve_status config_concatenation::get_resolve_status() const {
+        return resolve_status::UNRESOLVED;
+    }
+
+    shared_value config_concatenation::replace_child(shared_value const& child, shared_value replacement) const {
+        auto new_list = config_value::replace_child_in_list(_pieces, child, replacement);
+        if (new_list.empty()) {
+            return nullptr;
+        } else {
+            return make_shared<config_concatenation>(origin(), move(new_list));
+        }
+    }
+
+    bool config_concatenation::has_descendant(shared_value const &descendant) const {
+        return has_descendant_in_list(_pieces, descendant);
+    }
+
+    resolve_result<shared_value>
+    config_concatenation::resolve_substitutions(resolve_context const& context,
+                                                resolve_source const& source) const
+    {
+        // Right now there's no reason to pushParent here because the
+        // content of ConfigConcatenation should not need to replaceChild,
+        // but if it did we'd have to do this.
+        resolve_context new_context { context };
+
+        vector<shared_value> resolved;
+        resolved.reserve(_pieces.size());
+
+        for (auto& p : _pieces) {
+            // to concat into a string we have to do a full resolve,
+            // so unrestrict the context, then put restriction back afterward
+            auto restriction = new_context.restrict_to_child();
+            resolve_result<shared_value> result {new_context.unrestricted().resolve(p, source)};
+            auto r = result.value;
+            new_context = result.context.restrict(restriction);
+
+            if (r == nullptr) {
+                // it was optional... omit
+            } else {
+                resolved.push_back(r);
+            }
+        }
+
+        // now need to concat everything
+        vector<shared_value> joined { consolidate(resolved) };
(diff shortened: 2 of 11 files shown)


lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+            }
+        }
+
+        // now need to concat everything
+        vector<shared_value> joined { consolidate(resolved) };
+        // if unresolved is allowed we can just become another
+        // ConfigConcatenation
+        if (joined.size() > 1 && context.options().get_allow_unresolved()) {
+            return make_resolve_result(move(new_context), make_shared<config_concatenation>(origin(), joined));
+            //} else if (joined.empty()) {
```
MikaelSmith: This probably shouldn't be commented out.

whopper: Hah! Will have to check what happened here

lib/inc/internal/config_concatenation.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <internal/container.hpp>
+#include <string>
+#include <vector>
+
+namespace hocon {
+
+    class config_exception;
+
+    template<typename T>
+    struct resolve_result;
```
MikaelSmith: config_value forward-declares this, so you can skip doing it here.

lib/src/resolve_context.cc
```diff
@@ -22,4 +22,18 @@ namespace hocon {
...
+    }
+
+    resolve_context resolve_context::restrict(path restrict_to) const {
+        // TODO: implement
+        throw config_exception("resolve_context::restrict is not yet implemented");
+    }
+
+    resolve_context resolve_context::unrestricted() const {
+        // TODO: implement
+        throw config_exception("resolve_context::unrestricted is not yet implemented");
```
MikaelSmith: path is going to need an uninitialized state for this method. What ticket do we expect to finish these in? Ah, HC-42.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+#include <internal/resolve_result.hpp>
+#include <internal/resolve_source.hpp>
+#include <internal/resolve_context.hpp>
+
+using namespace std;
+
+namespace hocon {
+
+    config_concatenation::config_concatenation(shared_origin origin, std::vector<shared_value> pieces) :
+            config_value(move(origin)), _pieces(move(pieces)), _unmerged_values({ shared_from_this() }) {
```
MikaelSmith: Initializing here may not be safe. I'm not sure the shared_ptr exists when the constructor is called. I'd prefer to initialize the first time unmerged_values is called.

MikaelSmith: I feel like I'm overlooking a possibility here. The only use of `unmerged_values()` is in https://github.com/puppetlabs/cpp-hocon/blob/master/lib/src/values/config_value.cc#L223-L224. We could optimize for that somehow.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+                had_unmergeable = true;
+            }
+        }
+
+        if (!had_unmergeable) {
+            throw config_exception("Created concatenation without an unmergeable in it");
+        }
+    }
+
+    vector<shared_value> config_concatenation::pieces() const {
```
MikaelSmith: Why does this exist? I don't see it in the original.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+    resolve_status config_concatenation::get_resolve_status() const {
+        return resolve_status::UNRESOLVED;
+    }
+
+    shared_value config_concatenation::replace_child(shared_value const& child, shared_value replacement) const {
+        auto new_list = config_value::replace_child_in_list(_pieces, child, replacement);
+        if (new_list.empty()) {
+            return nullptr;
+        } else {
+            return make_shared<config_concatenation>(origin(), new_list);
```
MikaelSmith: Should `move(new_list)`.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+
+    bool config_concatenation::has_descendant(shared_value const &descendant) const {
+        return has_descendant_in_list(_pieces, descendant);
+    }
+
+    resolve_result<shared_value>
+    config_concatenation::resolve_substitutions(resolve_context const& context,
+                                                resolve_source const& source) const
+    {
+        // Right now there's no reason to pushParent here because the
```
MikaelSmith: The debugging might still be useful. But we haven't used or setup Leatherman.logging yet.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+        // content of ConfigConcatenation should not need to replaceChild,
+        // but if it did we'd have to do this.
+        resolve_source source_with_parent = source;
```
MikaelSmith: I think this can just be another const&, no need to make a copy. Which makes me think we can get rid of it and use `context`.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+                resolved.push_back(r);
+            }
+        }
+
+        // now need to concat everything
+        vector<shared_value> joined { consolidate(resolved) };
+        // if unresolved is allowed we can just become another
+        // ConfigConcatenation
+        if (joined.size() > 1 && context.options().get_allow_unresolved()) {
+            return make_resolve_result(move(new_context), make_shared<config_concatenation>(origin(), joined));
```
MikaelSmith: Probably want to `move(joined)`.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+
+    shared_value config_concatenation::concatenate(std::vector<shared_value> pieces) {
+        vector<shared_value> consolidated { consolidate(pieces) };
+        if (consolidated.empty()) {
+            return nullptr;
+        } else if (consolidated.size() == 1) {
+            return consolidated.front();
+        } else {
+            shared_origin merged_origin = simple_config_origin::merge_origins(consolidated);
+            return make_shared<config_concatenation>(merged_origin, consolidated);
```
MikaelSmith: `move(merged_origin)` and `move(consolidated)`.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+            return {};
```
MikaelSmith: This is never reached, should be removed.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+        }
+    }
+
+    // when you graft a substitution into another object,
+    // you have to prefix it with the location in that object
+    // where you grafted it; but save prefixLength so
+    // system property and env variable lookups don't get
+    // broken.
+    shared_value config_concatenation::relativized(std::string prefix) const {
+        vector<shared_value> new_pieces;
```
MikaelSmith: Good place to reserve.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+        for (auto& p : _pieces) {
+            new_pieces.push_back(p->relativized(prefix));
+        }
+
+        return make_shared<config_concatenation>(origin(), move(new_pieces));
+    }
+
+    bool config_concatenation::operator==(config_concatenation const& other) const {
+        // note that "origin" is deliberately NOT part of equality
+        return (_pieces == other._pieces);
```
MikaelSmith: This is fine, but not complete. I left out implementing config_value::operator== because it's messy (we probably need to implement config_value::operator== and have it call a virtual helper that children can override).

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+    bool config_concatenation::ignores_fallbacks() const {
+        // we can never ignore fallbacks because if a child ConfigReference
+        // is self-referential we have to look lower in the merge stack
+        // for its value.
+        return false;
+    }
+
+    void config_concatenation::render(std::string& result, int indent, bool at_root, config_render_options options) const {
+        for (auto& p : _pieces) {
+            p->render(result, indent, at_root, move(options));
```
MikaelSmith: Can't `move(options)`, it's a loop.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+        }
+    }
+
+    config_exception config_concatenation::not_resolved() const {
+        return config_exception("need to config#resolve(), see the API docs for config#resolve; substitution not resolved");
+    }
+
+    bool config_concatenation::is_ignored_whitespace(shared_value value) {
+        auto value_string = dynamic_pointer_cast<const config_string>(value);
+        return (value_string && !(value_string->was_quoted()));
```
MikaelSmith: You don't need the extra parens, but they don't hurt.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    void config_concatenation::join(std::vector<shared_value> & builder, shared_value right) {
```
MikaelSmith: Might be nice to bring over the comment: `Add left and right, or their merger, to builder.`

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,248 @@
...
+            // leave joined=null, cannot join
+        } else {
+            // handle primitive type or primitive type mixed with object or list
+            string s1 { left->transform_to_string() };
+            string s2 { right->transform_to_string() };
+            if (s1.empty() || s2.empty()) {
+                throw config_exception("Cannot concatenate object or list with a non-object-or-list: " +
+                    s1 + " and " + s2 + " are not compatible");
+            } else {
+                auto joined_origin = simple_config_origin::merge_origins(left->origin(), right->origin());
```
MikaelSmith: Should `move(joined_origin)`.

whopper:
@MikaelSmith woot! Took a pass to address most of your comments thus far and I *believe* unmerged_values is in order now.

As for the debugging bits, should I make a ticket to do a debugging pass once leatherman logging is in place?

(I'm going to have to start being way more std::move() happy :) )

MikaelSmith:
Yeah, ticket for it sounds good.

The case you've been missing is that we should generally std::move() things into constructor calls if we know we won't need them later.

lib/src/config_concatenation.cc
```diff
@@ -0,0 +1,246 @@
...
+            throw config_exception("Created concatenation without an unmergeable in it");
+        }
+    }
+
+    config_value_type config_concatenation::value_type() const {
+        throw not_resolved();
+    }
+
+    shared_ptr<vector<shared_value>> config_concatenation::unmerged_values() const {
+        return make_shared<vector<shared_value>>(vector<shared_value>{ shared_from_this() });
```
MikaelSmith: This seems redundant. You really needed both declarations?

lib/src/config_delayed_merge.cc
```diff
@@ -23,8 +23,8 @@ namespace hocon {
         throw config_exception("called value_type() on value with unresolved substitutions, need to config#resolve() first, see API docs");
     }
 
-    vector<shared_value> const& config_delayed_merge::unmerged_values() const {
-        return _stack;
+    shared_ptr<vector<shared_value>> config_delayed_merge::unmerged_values() const {
+        return make_shared<vector<shared_value>>(_stack);
```
MikaelSmith: This copies _stack. I was thinking we should represent _stack internally as a shared_ptr to vector. unmerged_values should return `shared_ptr<const vector<shared_value>>`.

MikaelSmith: Actually, since we're immediately copying the elements into another vector, let's look at returning by value and using move_iterators to move the elements. http://en.cppreference.com/w/cpp/iterator/move_iterator

lib/src/config_concatenation.cc
whopper: Heh yeah, that came from http://stackoverflow.com/questions/28877488/vector-construction-when-wrapped-in-a-shared-pointer when I couldn't get the initializer to work by itself.

whopper:
@MikaelSmith mk, updated with move iterators! Also hit the small comments.

PR #38, opened 2016-03-25
(HC-53) Implementing requirements for the config parser

MikaelSmith:


MikaelSmith:
@branan here's some pre-req work I've done. It might overlap with things you're doing.

lib/inc/internal/simple_include_context.hpp
```diff
@@ -6,15 +6,18 @@
 namespace hocon {
 
     class simple_include_context : public config_include_context {
+        using weak_parseable = std::weak_ptr<const parseable>;
     public:
-        simple_include_context(std::shared_ptr<parseable> parseable);
+        // Include context is part of a parseable, so it can only keep a weak pointer
+        // to the owning parseable.
+        simple_include_context(weak_parseable parseable);
```
MikaelSmith: Turns out this is only ever created in the constructor of parseable, and always passed a self reference. So I'm going to avoid this problem by passing in a reference instead. This is one of those places a Core Guidelines checker would be nice, to guarantee no one ever passes a bad reference.

lib/inc/hocon/config_value.hpp
```diff
@@ -52,6 +45,25 @@ namespace hocon {
         friend class config_concatenation;
     public:
         /**
+         * The type of a configuration value (following the <a
+         * href="http://json.org">JSON</a> type schema).
+         */
+        enum class type {
+            OBJECT, LIST, NUMBER, BOOLEAN, CONFIG_NULL, STRING, UNSPECIFIED
+        };
+        static char const* type_name(type t) {
```
branan: GCC 5.3 is throwing a "control reaches end of non-void function" warning here for me. Obviously that can't happen if `t` is a valid value. If I add a default that throws a `std::logic_error("Got impossible value for type enum")` it compiles

MikaelSmith: Ah, that sounds like a bug in the compiler (the switch covers all valid enum values for the enum class). I'll add a default section anyway.

MikaelSmith:
Hopefully fixed the compile error.

MikaelSmith:
Victory (I hope)! Finished enough to write a test case.

branan:
:+1: from me. This all looks right based on my reading of these parts of java-hocon, and it works for my simple test case of a non-resolver concatenation

MikaelSmith:
That seems like enough to merge it.

PR #40, opened 2016-04-01
(HC-54) Implement missing fallback/merge methods

MikaelSmith:
A number of fallback and merge overloads were missing, causing merges to
be skipped. Implement them.

### DIFF
--- a/lib/tests/concatenation_test.cc
+++ b/lib/tests/concatenation_test.cc
@@ -11,7 +11,10 @@ TEST_CASE("concatenation") {
         auto conf = parse_config(R"(a :  true "xyz" 123 foo)")->resolve();
         REQUIRE(conf->get_string("a") == "true xyz 123 foo");
     }
+}
 
+// These tests are expected to fail because resolve_result::resolve implementation is incomplete.
+TEST_CASE("concatenation pending", "[!shouldfail]") {
     SECTION("trivial string concatenation") {
         auto conf = parse_config(R"(a : ${x}foo, x = 1)")->resolve();
         REQUIRE(conf->get_string("a") == "1foo");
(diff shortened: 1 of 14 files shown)


lib/tests/concatenation_test.cc
```diff
@@ -12,14 +12,15 @@ TEST_CASE("concatenation") {
...
-    SECTION("trivial string concatenation") {
-        auto conf = parse_config(R"(a : ${x}foo, x = 1)")->resolve();
-        REQUIRE(conf->get_string("a") == "1foo");
-    }
-
-    SECTION("variable lookup o:") {
-        auto conf = parse_config(R"(a : ${x}, x = 1)")->resolve();
-        REQUIRE(conf->get_string("a") == "1");
-    }
+// Tests fail now because resolve_result::resolve implementation is incomplete.
```
branan: What's the failure you're seeing? I was going to dive into finishing up resolve today/tomorrow, so I kind of want to get that in so we can uncomment these

branan: I may just base my work on this PR and sort it out myself

branan:
I think we'll probably catch up on the coverage loss as we add more tests, so I'm fine letting that slip.

lib/tests/concatenation_test.cc
MikaelSmith: Hmm, I'm not sure. I may have fixed it later. I'll restore the tests.

branan: Ah, yup. That's one of the things I was going to try to tackle today/tomorrow. I guess we can just merge this with those tests disabled, and I'll uncomment them when I get that stuff fixed

branan: Does catch have a better way of flagging tests pending than commenting them out?

MikaelSmith: Yes, https://github.com/philsquared/Catch/blob/master/docs/test-cases-and-sections.md#special-tags. I'll update to use `[!shouldfail]`.

MikaelSmith: Updated, hopefully it works in CI.

PR #41, opened 2016-04-04
(HC-54) More fixes for tests

MikaelSmith:

PR #42, opened 2016-04-05
(HC-42) Add memoization support to resolve_context

branan:
This is gross and probably super inefficient (lots of copying
unordered_maps around), but mirrors the original pretty closely.

### DIFF
--- a/lib/inc/internal/resolve_context.hpp
+++ b/lib/inc/internal/resolve_context.hpp
@@ -4,6 +4,8 @@
 #include <hocon/config_resolve_options.hpp>
 #include <hocon/path.hpp>
 
+#include <unordered_map>
+
 namespace hocon {
 
     class resolve_source;
@@ -27,7 +29,22 @@ namespace hocon {
         static shared_value resolve(shared_value value, shared_object root, config_resolve_options options);
 
     private:
+        struct memo_key {
+            shared_value value;
+            path restrict_to_child;
+            bool operator==(const memo_key& other) const {
+                return value == other.value && restrict_to_child == other.restrict_to_child;
+            }
+        };
+
+        struct memo_key_hash {
+            std::size_t operator()(const memo_key&) const;
+        };
+        using resolve_memos = std::unordered_map<memo_key, shared_value, memo_key_hash>;
         config_resolve_options _options;
         path _restrict_to_child;
+        resolve_memos _memos;
+
+        resolve_context memoize(const memo_key& key, const shared_value& value) const;
     };
 }  // namespace hocon
--- a/lib/src/resolve_context.cc
+++ b/lib/src/resolve_context.cc
@@ -23,16 +23,49 @@ namespace hocon {
 
     resolve_result<shared_value> resolve_context::resolve(shared_value original, resolve_source const& source) const
     {
-        auto result = original->resolve_substitutions(*this, source);
-        auto resolved = result.value;
+        memo_key full_key {original, {}};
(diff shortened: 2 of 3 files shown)


lib/src/resolve_context.cc
```diff
@@ -23,16 +23,47 @@ namespace hocon {
...
+            auto cached_iter = _memos.find(full_key);
+            if (cached_iter != _memos.end()) {
+                cached = cached_iter->second;
+            }
+        }
+        
+        if (cached) {
+            return make_resolve_result(*this, cached);
+        } else {
+            auto result = original->resolve_substitutions(*this, source);
```
MikaelSmith: Mind adding a note about still needing to handle cycles?

lib/src/resolve_context.cc
```diff
@@ -23,16 +23,47 @@ namespace hocon {
...
+            auto resolved = result.value;
+            auto with_memo = result.context;
+
+            if (!resolved || resolved->get_resolve_status() == resolve_status::RESOLVED) {
+                with_memo = with_memo.memoize(full_key, resolved);
+            }  else {
+                if (is_restricted_to_child()) {
+                    if (restricted_key.value == nullptr && restricted_key.restrict_to_child.empty()) {
```
MikaelSmith: This doesn't seem to follow the original logic. Why?

lib/inc/internal/resolve_context.hpp
```diff
@@ -27,7 +29,22 @@ namespace hocon {
...
+
+        struct memo_key_hash {
+            std::size_t operator()(const memo_key&) const;
+        };
+        using resolve_memos = std::unordered_map<memo_key, shared_value, memo_key_hash>;
         config_resolve_options _options;
         path _restrict_to_child;
+        resolve_memos _memos;
+
+        resolve_context memoize(const memo_key& key, const shared_value& value);
```
MikaelSmith: Looks like this could be const.

lib/src/resolve_context.cc
```diff
@@ -23,16 +23,47 @@ namespace hocon {
...
+                cached = cached_iter->second;
+            }
+        }
+        
+        if (cached) {
+            return make_resolve_result(*this, cached);
+        } else {
+            auto result = original->resolve_substitutions(*this, source);
+            auto resolved = result.value;
+            auto with_memo = result.context;
```
MikaelSmith: Might want to make this a ref, so we're not copying the more often than we need to.

lib/src/resolve_context.cc
branan: The original logic uses a `NULL` hash key to indicate it's not initialized. Since I'm treating the key as a value type, I'm just checking if it's an empty/default initialization

MikaelSmith: Ah, ok.

lib/src/resolve_context.cc
branan: done

lib/inc/internal/resolve_context.hpp
branan: done

lib/src/resolve_context.cc
branan: done

branan:
@MikaelSmith feedback addressed, and rebased on top of #41 to ensure there were no behavioral conflicts.

MikaelSmith:
Linting :(

branan:
Pushed update for the lints

PR #43, opened 2016-04-06
Hc 55 concatenation tests

branan:
This implements all of the string concatenation tests, and some of the
array and object tests. The full test suite requires us to have all
the accessors for lists and objects implemented, which we don't
yet. Since we're at the end of a sprint, getting this subset done
seemed expedient.

### DIFF
--- a/lib/tests/concatenation_test.cc
+++ b/lib/tests/concatenation_test.cc
@@ -1,6 +1,7 @@
 #include <catch.hpp>
 
 #include <hocon/config.hpp>
+#include <hocon/config_exception.hpp>
 #include "test_utils.hpp"
 
 using namespace hocon;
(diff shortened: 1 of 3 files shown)


lib/tests/concatenation_test.cc
```diff
@@ -18,9 +19,321 @@ TEST_CASE("concatenation") {
...
+    }
+
+    SECTION("concatenate two undefined substitutions with an empty string") {
+        auto conf = parse_config(R"(a = ""${?foo}${?bar})")->resolve();
+        REQUIRE(conf->get_string("a") == "");
+    }
+}
+
+TEST_CASE("Concatenation pending '+=' implementation", "[!shouldfail]") {
+    // TODO: emptyArrayPlusEquals
```
MikaelSmith: These are expected to fail because of rendering in exceptions?

branan: All += stuff fails, it's totally unimplemented. https://github.com/puppetlabs/cpp-hocon/blob/master/lib/src/config_parser.cc#L136

Probably needs Yet Another Ticket

PR #44, opened 2016-04-06
(HC-52) Add basic config substitution tests

HAIL9000:
Add some basic tests for substitution and the relevant helper methods.
Also do a bit of code cleanup including making the needed operators
public and replacing some global instance variables with functions
instead (to avoid and error where the variable is not initialized
at the time the tests are run).

### DIFF
--- a/lib/src/tokenizer.cc
+++ b/lib/src/tokenizer.cc
@@ -185,10 +185,16 @@ namespace hocon {
     }
 
     /** Characters JSON allows a number to start with */
-    static const string first_number_chars = "0123456789-";
+    static string first_number_chars() {
+        static const string first_number_chars_ = "0123456789-";
+        return first_number_chars_;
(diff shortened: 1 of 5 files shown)


lib/src/tokenizer.cc
```diff
@@ -185,10 +185,16 @@ namespace hocon {
     }
 
     /** Characters JSON allows a number to start with */
-    static const string first_number_chars = "0123456789-";
+    static string first_number_chars() {
```
MikaelSmith: This is done to avoid too much global static initialization (and some races we got into using static initialization for test variables). @peterhuene @kylog @branan have patterns you prefer here? Singletons could also be used, but I don't have a strong case for it.

branan: I'm fine with lazy statics :+1:

PR #47, opened 2016-05-26
(HC-56) Implement cycle detection in resolve implementation

whopper:
Also port cycle-related tests in config_substitution tests.

### DIFF
--- a/lib/src/resolve_context.cc
+++ b/lib/src/resolve_context.cc
@@ -3,13 +3,17 @@
 #include <internal/resolve_context.hpp>
 #include <internal/resolve_result.hpp>
 #include <internal/resolve_source.hpp>
+#include <algorithm>
 
 using namespace std;
 
 namespace hocon {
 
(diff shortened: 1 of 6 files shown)


lib/src/resolve_context.cc
```diff
@@ -3,13 +3,17 @@
...
 using namespace std;
 
 namespace hocon {
 
+    resolve_context::resolve_context(config_resolve_options options, path restrict_to_child, vector<shared_value> cycle_markers)
+         : _options(move(options)), _restrict_to_child(move(restrict_to_child)), _cycle_markers(move(cycle_markers)) { }
+
     resolve_context::resolve_context(config_resolve_options options, path restrict_to_child)
-         : _options(move(options)), _restrict_to_child(move(restrict_to_child)) { }
+         : resolve_context(move(options), move(restrict_to_child), move(vector<shared_value> {})) { }
```
MikaelSmith: Just a note, `move` isn't necessary here because you're initializing the value, you get automatic copy elision.

PR #52, opened 2016-06-01
(HC-58) Implement lookup of variable substitutions in the system environment

branan:


whopper:
👍 TIL about leatherman::util::environment::each

PR #55, opened 2016-06-07
(HC-57) Add render for config_value family

HAIL9000:
Implement the render method for the family of classes that inherit
from the config_value class.

### DIFF
--- a/lib/src/values/simple_config_object.cc
+++ b/lib/src/values/simple_config_object.cc
@@ -9,6 +9,8 @@
 
 #include <algorithm>
 #include <unordered_set>
+#include <boost/algorithm/string.hpp>
+#include <internal/tokens.hpp>
 
 using namespace std;
 
@@ -176,7 +178,15 @@ namespace hocon {
     }
 
     bool simple_config_object::operator==(config_value const& other) const {
-        return equals<simple_config_object>(other, [&](simple_config_object const& o) { return _value == o._value; });
+        return equals<simple_config_object>(other, [&](simple_config_object const& o) {
+            if (_value.size() != o._value.size()) { return false; }
+
+            bool still_equal = true;
+            for (auto pair : _value) {
+                still_equal = *(o._value.at(pair.first)) == *(_value.at(pair.first));
+            }
+            return still_equal;
+        });
     }
 
     resolve_result<shared_value>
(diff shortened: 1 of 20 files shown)


lib/src/values/simple_config_list.cc
```diff
@@ -99,7 +99,9 @@ namespace hocon {
             }
 
             if (equal(begin(), end(), o.begin(),
-                      [](shared_value const& a, shared_value const& b) { return a == b; })) {
+                      [](shared_value const& a, shared_value const& b) {
```
MikaelSmith: I think you misunderstood the function here. It checks if the pointers are equal, then it does a deep comparison below if that failed.

lib/tests/path_test.cc
```diff
@@ -31,6 +31,7 @@ TEST_CASE("path to string", "[path]") {
 
 void render_test(path p, string expected) {
     REQUIRE(p.render() == expected);
+    auto parsed = path_parser::parse_path(expected);
```
MikaelSmith: I don't think you meant to leave this in.

lib/src/values/simple_config_object.cc
```diff
@@ -390,4 +400,104 @@ namespace hocon {
         }
     }
 
+    bool compare(string& a, string& b) {
```
MikaelSmith: http://en.cppreference.com/w/cpp/algorithm/sort notes that this function should use `string const&` arguments. Specifically, the signature is `bool cmp(const Type1 &a, const Type2 &b);`. That's probably what's causing the compile error.

HAIL9000:
@MikaelSmith Comments addressed!

MikaelSmith:
👍

lib/tests/conf_parser_test.cc
```diff
@@ -47,13 +47,13 @@ TEST_CASE("valid conf works") {
...
         // let's also check round-trip rendering
-        //auto rendered = our_ast->render();
-        //CAPTURE(rendered);
-        //auto reparsed = parse(rendered);
-        //REQUIRE(*our_ast == *reparsed);
+        auto rendered = our_ast->render();
+        CAPTURE(rendered);
+        auto reparsed = parse(rendered);
+        auto foo = *our_ast == *reparsed;
+        REQUIRE(foo);
```
branan: "foo"?

branan:
Other than squashing that REQUIRE down to one line (or giving it a better variable name) I'm :+1:

PR #57, opened 2016-06-08
(maint) Fixed a couple of misused moves.

Magisus:
When attempting to compile on my system, both of these moves caused the error "moving a temporary object prevents copy elision." After removing the offending moves, the project compiles and all the tests pass.


whopper:
@Magisus Oh one thing, cpp-hocon doesn't happen to have a commit message checker but other projects expect the `maint` to be in parens `(maint)`

Magisus:
Whoops, I'll fix it.

whopper:
👍 will merge when travis goes green again!

PR #59, opened 2016-06-09
(maint) Mark pending tests in title

HAIL9000:
Add "(pending)" to test cases expected to fail to make the ctest
output more redable.


Magisus:
This will be nice. We should definitely try to be consistent about doing this in the future if we know the tests are failing because something isn't implemented correctly yet. Merging!

PR #60, opened 2016-06-10
(HC-66) Implement create_value_under_path()

Magisus:
This implements create_value_under_path() in the config_parser. Tests that previously failed due to this method not existing now either pass or fail due to unrelated bugs (see HC-78).

We should decide on the best format for marking known test failures. I used "(pending <ticket_number>)" where a ticket had been made, and just "(pending)" otherwise. But this is not consistent throughout at the moment.

### DIFF
--- a/lib/src/config_parser.cc
+++ b/lib/src/config_parser.cc
@@ -83,6 +83,37 @@ namespace hocon { namespace config_parser {
         return v;
     }
 
+    shared_object parse_context::create_value_under_path(path p, shared_value value) {
+        // for path foo.bar, we are creating { "foo" : { "bar" : value } }
+        vector<shared_string> keys;
+
+        shared_string key = p.first();
+        path remaining = p.remainder();
+        while (key != nullptr) {
+            keys.push_back(key);
+            if (remaining.empty()) {
+                break;
+            } else {
+                key = remaining.first();
+                remaining = remaining.remainder();
+            }
+        }
+
+        auto current = keys.end();
+        current--;
+        auto new_value = unordered_map<string, shared_value>({ {**current, value} });
+        shared_object obj = make_shared<simple_config_object>(value->origin()->with_comments(vector<string>{}),
+                                                              new_value);
+
+        while (current != keys.begin()) {
+            current--;
+            new_value = unordered_map<string, shared_value>({ {**current, obj} });
+            obj = make_shared<simple_config_object>(value->origin()->with_comments(vector<string>{}), new_value);
+        }
+
+        return obj;
+    }
+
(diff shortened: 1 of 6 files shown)


whopper:
@Magisus woot! Would you mind rebasing / squashing your commits into one or two (or however many make sense) commits with `(HC-66)` in their titles? Our commit message linter actually doesn't care about the GitHub PR title, but rather the individual commit messages.

lib/src/config.cc
```diff
@@ -135,6 +135,8 @@ namespace hocon {
                                                  path original_path) {
         shared_value v = self->peek_assuming_resolved(key, original_path);
         if (!v) {
+            // TODO: remove, debugging
```
whopper: Looks like this is meant to be removed

lib/src/config_parser.cc
```diff
@@ -254,7 +285,15 @@ namespace hocon { namespace config_parser {
...
                     }
 
-                    throw bug_or_broken_exception("createValueUnderPath not implemented");
+                    shared_object obj = create_value_under_path(remaining, new_value);
+                    auto existing = values.find(*key);
+                    if (existing != values.end()) {
+                        obj = dynamic_pointer_cast<const config_object>(obj->with_fallback(existing->second));
+                        assert(obj);
+                        existing->second = obj;
+                    } else {
```
whopper: I don't believe this should be an else clause, from the looks of the Java source:

```
                        if (existing != null) {
                            obj = obj.withFallback(existing);
                        }
                        values.put(key, obj);
```

Magisus: That confused me too, but doing it just like the java was actually a bug. For some reason the value was not replacing properly from unordered_map::insert, so instead I have two cases: update the object already stored in the map, or add a new object to the map. I copied an example of this I found elsewhere in that file.

whopper: @Magisus Oh, interesting! Well in that case, 👍 thanks for saving us a potential bug :D

Magisus: And for future reference, the docs are making it sounds like if the key is already in the map, insert() just does nothing. Whereas in Java it always overwrites.

whopper: Ah, perhaps we should use `map::operator[]` instead, which I think will add a new key if it doesn't exist and overwrite otherwise: `map[key] = value;` http://stackoverflow.com/a/8801171

Magisus: Oh I like that, fixing both locations.

whopper:
👍 lgtm! @Magisus Any final additions before I merge?

Magisus:
I don't think so. Go for it :)

PR #61, opened 2016-06-15
(HC-43) Implement unwrap functions for config values

Magisus:
This implements unwrapped() methods for all the various config values. For some, this involves recursive unwrapping of objects and lists, and some others throw "not resolved" exceptions.

Having this method allowed me to implement some additional tests that were blocked simply by unwrapped(). However, some nearby tests rely on the specifically-typed config::get_*_list methods. Do we consider implementing those part of this ticket, or should we leave it for another time?


branan:
I think those can be handled as a separate ticket (which we should file if it doesn't exist yet)

Magisus:
Looks like we do already have one, HC-75.

PR #62, opened 2016-06-22
(HC-36) Implement ConfigValueFactory API and tests

Magisus:
This class is used to turn C++ types into config_values. The types it
can accept are limited by the bounded types of the boost::variant used
when unwrapping config_values.

### DIFF
--- a/lib/src/config_value_factory.cc
+++ b/lib/src/config_value_factory.cc
@@ -1,0 +1,68 @@
+#include <hocon/config_value_factory.hpp>
+#include <internal/values/config_null.hpp>
+#include <internal/values/config_string.hpp>
+#include <internal/values/config_long.hpp>
+#include <internal/values/config_double.hpp>
+#include <internal/values/config_int.hpp>
+#include <internal/values/config_boolean.hpp>
+#include <internal/values/simple_config_list.hpp>
+#include <internal/values/simple_config_object.hpp>
+
+namespace hocon {
+
+    using namespace std;
+
+    class config_value_visitor : public boost::static_visitor<shared_value> {
+    public:
+        // TODO: If use cases of from_any_ref require other types to produce config_nulls,
+        // we can revise this behavior
+        shared_value operator()(boost::blank null_value) const {
+            return make_shared<const config_null>(nullptr);
+        }
+
+        shared_value operator()(string str) const {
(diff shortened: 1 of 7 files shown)


lib/src/config_value_factory.cc
```diff
@@ -0,0 +1,63 @@
...
+    public:
+        shared_value operator()(boost::blank null_value) const {
+            return make_shared<const config_null>(nullptr);
+        }
+
+        shared_value operator()(string str) const {
+            return make_shared<const config_string>(nullptr, str, config_string_type::QUOTED);
+        }
+
+        // TODO: Java passes null for original text, does this do the right equivalent thing?
```
MikaelSmith: The place to look is where we use that value, aka https://github.com/puppetlabs/cpp-hocon/blob/master/lib/src/values/config_long.cc#L10-L17.

lib/tests/config_value_factory_test.cc
```diff
@@ -0,0 +1,63 @@
...
+
+    SECTION("false should convert to config_boolean") {
+        auto value = config_value_factory::from_any_ref(false, "");
+        REQUIRE(dynamic_pointer_cast<const config_boolean>(value));
+        REQUIRE(boost::get<bool>(value->unwrapped()) == false);
+    }
+
+    // TODO: Where do config_nulls usually come from? What type might conceivably be converted to them in C++?
+    // Because passing a boost::blank causes one to be produced, but in theory we might want other values to do so,
+    // but because of value semantics, nullptr is out.
```
MikaelSmith: I think this happens if you have a reference that can't be resolved. You could construct a config example to test that out.

lib/src/config_value_factory.cc
Magisus: This does in fact produce the correct behavior. Will remove the TODO.

lib/tests/config_value_factory_test.cc
Magisus: The from_any_ref method is not used internally anywhere, and in all the tests, the only value passed to test config_null is 'null', with no indication of how that 'null' was produced. I feel like we will have to see what our use cases for this method are before we can decide what C++ values should produce config_nulls.

In every case that I have come across, failure to resolve results in an exception being thrown.

PR #64, opened 2016-07-14
(FACT-1456) Add utility to wrap hocon in boost::program_options

branan:
This adds a `hocon::program_options` namespace with a `parse_hocon`
function. This function can be used the same as any of the normal
`program_options` parsers to generate config values, except it takes
as its input a hocon `shared_config` object.

HOCON allows representing certain data structures that
`boost::program_options` can not - specifically hashes inside arrays
cause trouble. We try to reprsent these when possible, (using a syntax
that treats the array index as a "key"), but defining actual
`po::options_description` sets that can represent these is Hard (if
not impossible).

### DIFF
--- a/lib/inc/hocon/program_options.hpp
+++ b/lib/inc/hocon/program_options.hpp
@@ -1,0 +1,78 @@
+#pragma once
+
+#include "config.hpp"
+#include "config_list.hpp"
+
+#include <boost/format.hpp>
+#include <boost/program_options.hpp>
+
+namespace hocon { namespace program_options {
+    namespace po = boost::program_options;
+
+    template<class charT>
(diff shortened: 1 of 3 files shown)


Magisus:
I really like this. It will make integration with existing command line options fairly painless and allow us to utilize some of our existing conflict-prevention logic (though depending on the structure of the config file, nesting may make this slightly more complicated). 

This seems particularly suited to grabbing the global setup options, which may end up all in one HOCON object and most likely will not have complicated structure. The different input types these methods accept will give us the flexibility to parse just parts of the HOCON config if that seems appropriate.

MikaelSmith:
Oh lint.

PR #65, opened 2016-07-15
(HC-75) Implement config::get_*_list methods.

Magisus:
This commit implements the various get_list methods for supported types,
using a templated backing function which will throw exceptions if the
list is not homogeneous and of the desired type.

### DIFF
--- a/lib/inc/hocon/config.hpp
+++ b/lib/inc/hocon/config.hpp
@@ -6,6 +6,7 @@
 #include "config_object.hpp"
 #include "config_resolve_options.hpp"
 #include "config_value.hpp"
+#include "config_list.hpp"
 #include "config_exception.hpp"
 #include "path.hpp"
 #include <vector>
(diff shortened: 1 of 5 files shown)


lib/inc/hocon/config.hpp
```diff
@@ -570,6 +570,9 @@ namespace hocon {
         virtual unwrapped_value get_any_ref(std::string const& path) const;
         virtual std::shared_ptr<const config_value> get_value(std::string const& path) const;
 
+        template<typename T>
+        std::vector<T> get_homogeneous_unwrapped_list(std::string const& path) const;
```
MikaelSmith: Why make this public? I don't see any public use of it, and to create new templated forms the full function would need to be in the header file.

Magisus: I was actually planning to use it as part of the access layer I'm writing for the Facter stuff, to allow just writing one template function there instead of calling into all those explicitly named functions that HOCON is supposed to provide. But I do realize I've done something wrong, because I'm fighting the linker again, as usual xD

PR #67, opened 2016-07-19
(maint) Fixup pending tests that actually pass

branan:


Magisus:
Looks good, all tests either pass or fail as expected on my machine. If checks are green, I'll merge.

PR #69, opened 2016-07-26
(HC-85) Fix various platform-dependent build errors

Magisus:
This does not account for the errors on Windows, which had to be fixed in puppet-agent, or on OSX, which are pending a tooling update.

### DIFF
--- a/lib/src/path_parser.cc
+++ b/lib/src/path_parser.cc
@@ -206,7 +206,7 @@ namespace hocon {
             return true;
         }
 
-        for (signed char c : s) {
+        for (char c : s) {
             if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '_') {
                 last_was_dot = false;
                 continue;
--- a/lib/src/path_parser.cc
+++ b/lib/src/path_parser.cc
@@ -206,7 +206,7 @@ namespace hocon {
             return true;
         }
 
-        for (char c : s) {
+        for (signed char c : s) {
             if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '_') {
                 last_was_dot = false;
                 continue;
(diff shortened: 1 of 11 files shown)


lib/tests/config_node_test.cc
```diff
@@ -158,7 +158,7 @@ void top_level_value_replace_test(shared_node_value value,
 TEST_CASE("replace nodes", "[config-node]") {
     SECTION("simple values can be replaced") {
         top_level_value_replace_test(int_node(10), int_node(15));
-        top_level_value_replace_test(long_node(numeric_limits<int>::max() + 1), int_node(1));
+        top_level_value_replace_test(long_node(int64_t(numeric_limits<int>::max()) + 1), int_node(1));
```
MikaelSmith: Please use a static_cast here, as in `static_cast<int64_t>(numeric_limits<int>::max())`.

lib/src/path_parser.cc
```diff
@@ -206,7 +206,7 @@ namespace hocon {
             return true;
         }
 
-        for (char c : s) {
+        for (signed char c : s) {
```
MikaelSmith: This one doesn't make any sense. String always have `char`, not `signed char`.

lib/src/tokenizer.cc
```diff
@@ -22,7 +22,7 @@ namespace hocon {
     /** Whitespace Saver */
     token_iterator::whitespace_saver::whitespace_saver() : _last_token_was_simple_value(false) { }
 
-    void token_iterator::whitespace_saver::add(char c) {
+    void token_iterator::whitespace_saver::add(signed char c) {
```
MikaelSmith: This should really be reworked to use `istream` directly, it already supports the idea of `putback`. However, that can be left as a separate ticket if we need to get things building.

Magisus:
Still passing on everything besides OSX, after responding to Michael's feedback.

MikaelSmith:
👍

PR #70, opened 2016-07-28
(HC-84) Improve exception messages from config API

Magisus:
This updates the config class to use more detailed exception types and
correctly propagate useful messages.

### DIFF
--- a/lib/inc/hocon/config_exception.hpp
+++ b/lib/inc/hocon/config_exception.hpp
@@ -45,9 +45,9 @@ namespace hocon {
      * was set to null.
      */
     struct null_exception : public missing_exception {
-        null_exception(config_origin const& origin, std::string const& path, std::string const& expected) :
-                missing_exception(origin, (expected.empty() ? "Configuration key '" + path + "' is null"
-                                                            : "Configuration key '" + path + "' is set to null but expected " + expected)) { }
+        null_exception(config_origin const& origin, std::string const& path, std::string const& expected = "") :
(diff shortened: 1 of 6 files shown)


lib/inc/hocon/config_exception.hpp
```diff
@@ -45,9 +45,9 @@ namespace hocon {
      * was set to null.
      */
     struct null_exception : public missing_exception {
-        null_exception(config_origin const& origin, std::string const& path, std::string const& expected) :
-                missing_exception(origin, (expected.empty() ? "Configuration key '" + path + "' is null"
-                                                            : "Configuration key '" + path + "' is set to null but expected " + expected)) { }
+        null_exception(config_origin const& origin, std::string const& path, std::string const& expected = "") :
+                missing_exception(origin, (expected.empty() ? "Configuration key \"" + path + "\" is null"
```
Magisus: Due to the confusion we had surrounding single quotes, I found this message to be misleading.

lib/src/config.cc
```diff
@@ -377,14 +385,5 @@ namespace hocon {
         return make_shared<simple_config_object>(origin, move(values), resolve_status::RESOLVED, false);
     }
 
-    not_resolved_exception config::improve_not_resolved(path what, not_resolved_exception const& original) {
```
Magisus: Moved to resolve_source, which is the only place it is used.

MikaelSmith:
👍

PR #71, opened 2016-08-12
(HC-62) Implement duration parsing

Magisus:
This implements parsing durations with the aid of boost::chrono::durations. There is a lot of converting back and forth between them, but this seemed the easiest way to handle the type system, short of writing my own custom type.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ cmake_minimum_required(VERSION 3.2.2)
 # Project Setup - modify to match project naming
 ## Source code for a simple command-line executable for a dynamic library will be generated from the project name.
 ## The command-line and library names will be based off the project name.
-project(cpp-hocon VERSION 0.1.0)
+project(cpp-hocon VERSION 0.1.2)
 
 string(MAKE_C_IDENTIFIER ${PROJECT_NAME} PROJECT_C_NAME)
 string(TOUPPER ${PROJECT_C_NAME} PROJECT_NAME_UPPER)
--- a/lib/inc/hocon/config.hpp
+++ b/lib/inc/hocon/config.hpp
@@ -17,6 +17,8 @@
 
 namespace hocon {
 
+    enum class time_unit { NANOSECONDS, MICROSECONDS, MILLISECONDS, SECONDS, MINUTES, HOURS, DAYS };
+
     /**
      * An immutable map from config paths to config values. Paths are dot-separated
      * expressions such as <code>foo.bar.baz</code>. Values are as in JSON
@@ -595,7 +597,26 @@ namespace hocon {
         virtual std::vector<shared_object> get_object_list(std::string const& path) const;
         virtual std::vector<shared_config> get_config_list(std::string const& path) const;
 
-        // TODO: memory and duration parsing
+        // TODO: memory parsing
+
+        /**
+         * Gets a value as a decimal number of the specified units.
+         * Correctly handles durations within the range +/-2^63 seconds.
+         * @param path the path to the time value
+         * @param unit the units of the number returned
+         * @return a double representing the value converted to the requested units
+         */
+        virtual double get_duration_as_double(std::string const& path, time_unit unit) const;
+
+        /**
+         * Gets a value as an integer number of the specified units.
+         * If the result would have a fractional part, the number is truncated.
+         * Correctly handles durations within the range +/-2^63 seconds.
+         * @param path the path to the time value
+         * @param unit the units of the number returned
+         * @return a 64-bit integer representing the value converted to the requested units
+         */
+        virtual int64_t get_duration_as_long(std::string const& path, time_unit unit) const;
 
         /**
          * Clone the config with only the given path (and its children) retained;
@@ -681,6 +702,23 @@ namespace hocon {
         // TODO: memory and duration parsing
 
     private:
+        /**
+         * Parses a duration string. If no units are specified in the string, it is assumed to be in
+         * milliseconds.
+         * Valid suffixes include ns, us, ms, s, m, h, and d. The units can also be specified as complete
+         * words (e.g. "seconds").
+         * @param input the string to parse
+         * @param origin_for_exception the origin of the value being parsed
+         * @param path_for_exception the path to the value being parsed
+         * @return the value parsed as a time_duration
+         */
+        static duration parse_duration(std::string input, shared_origin origin_for_exception, std::string path_for_exception);
+
+        static duration convert(int64_t number, time_unit units);
+        static duration convert(double number, time_unit units);
+        static time_unit get_units(std::string const& unit_string);
+        duration get_duration(std::string const& path) const;
+
         shared_value has_path_peek(std::string const& path_expression) const;
         shared_value peek_path(path desired_path) const;
 
--- a/lib/src/config.cc
+++ b/lib/src/config.cc
@@ -7,12 +7,20 @@
 #include <internal/values/config_boolean.hpp>
 #include <internal/values/config_null.hpp>
 #include <internal/values/config_number.hpp>
+#include <internal/values/config_double.hpp>
+#include <internal/values/config_long.hpp>
+#include <internal/values/config_int.hpp>
 #include <internal/values/config_string.hpp>
 #include <internal/values/simple_config_object.hpp>
 #include <internal/parseable.hpp>
 #include <internal/simple_includer.hpp>
+
+#include <boost/lexical_cast.hpp>
+#include <boost/algorithm/string/trim.hpp>
 #include <leatherman/util/environment.hpp>
 #include <leatherman/locale/locale.hpp>
+
+#include <cfenv>
 
 // Mark string for translation (alias for leatherman::locale::format)
 using leatherman::locale::_;
@@ -195,7 +203,7 @@ namespace hocon {
         return v->value_type() == config_value::type::CONFIG_NULL;
     }
 
-    shared_ptr<const config_value> config::get_value(string const& path_expression) const {
+    shared_value config::get_value(string const& path_expression) const {
         return find(path_expression, config_value::type::UNSPECIFIED);
     }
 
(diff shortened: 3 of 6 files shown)


CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ cmake_minimum_required(VERSION 3.2.2)
 # Project Setup - modify to match project naming
 ## Source code for a simple command-line executable for a dynamic library will be generated from the project name.
 ## The command-line and library names will be based off the project name.
-project(cpp-hocon VERSION 0.1.0)
+project(cpp-hocon VERSION 0.1.2)
```
Magisus: This got missed when we tagged. When is the correct time to fix it, if not part of this PR?

Magisus:
@branan You made a comment about not wanting to return double values of nanosecods, favoring integer values of milliseconds. How do you feel about throwing out support for parsing durations with decimal points entirely? This would simplifying things both from the HOCON point of view and allow us to use built in types of boost::chrono::duration. In other words, if you want fractions of an hour, you have to go down to minutes in your config file.

If not, it looks like `duration` does not support double-to-int type conversions. So we might have to do some math explicitly.

lib/src/config.cc
```diff
@@ -304,6 +311,88 @@ namespace hocon {
...
+        try {
+            number = boost::lexical_cast<double>(number_string);
+        } catch (boost::bad_lexical_cast& ex) {
+            throw bad_value_exception(*origin_for_exception, path_for_exception, "Value '" + number_string + "' could not be converted to a number.");
+        }
+        if (unit_string.empty() || unit_string == "ms" || unit_string == "millis" || unit_string == "milliseconds") {
+            return milliseconds(static_cast<int64_t>(number));
+        } else if (unit_string == "d" || unit_string == "days") {
+            // boost::chrono::duration does not support conversion from doubles to ints very well
+            int64_t millis = static_cast<int64_t>(number * 86400000);
```
branan: should this be unsigned? I don't think negative durations are meaningful...

branan:
At this point are we using `boost::chrono` as anything but "a way to provide some type safety around durations"? If that's all, should we maybe just return a `uint64_t` to keep our interface simpler to use directly with other libraries that aren't boost-based?

lib/src/config.cc
Magisus: Technically HOCON supports negatives. There is a test for it. But I suppose that's up to use.

lib/src/config.cc
```diff
@@ -7,14 +7,21 @@
...
 #include <internal/values/config_number.hpp>
+#include <internal/values/config_double.hpp>
+#include <internal/values/config_long.hpp>
+#include <internal/values/config_int.hpp>
 #include <internal/values/config_string.hpp>
 #include <internal/values/simple_config_object.hpp>
 #include <internal/parseable.hpp>
 #include <internal/simple_includer.hpp>
 
+#include <boost/regex.hpp>
```
MikaelSmith: regex looks like it's no longer used.

lib/src/config.cc
```diff
@@ -304,6 +311,88 @@ namespace hocon {
         return long_list;
     }
 
+    string config::get_units(string s) {
+        int i = s.length() - 1;
```
MikaelSmith: This seems like it could be a trim_right call.

lib/src/config.cc
```diff
@@ -304,6 +311,88 @@ namespace hocon {
...
+            throw bad_value_exception(*origin_for_exception, path_for_exception, "No number in duration value '" + input + "'");
+        }
+
+        if (unit_string.length() > 2 && unit_string.back() != 's') {
+            unit_string += "s";
+        }
+
+        double number = 0;
+        try {
+            number = boost::lexical_cast<double>(number_string);
```
MikaelSmith: This can lose precision in a very minor case: x86 systems (usually) support extended precision floating point, meaning they use 80-bits to represent doubles - 1 bit for sign, 15 bits for the exponent, 1 bit for whether it's '0.xxx' or '1.xxx', and 63 bits for the fraction.

So this is safe if we only support 63-bits of precision (int64_t), but not 64-bits (uint64_t). The following example shows the difference.

```
#include <boost/lexical_cast.hpp>
#include <string>
#include <iostream>

int main(int argc, char** argv)
{
    std::string s = "18446744073709551615";
    double number = boost::lexical_cast<double>(s);
    uint64_t integer = static_cast<uint64_t>(number);
    std::cout << s << " != " << integer << std::endl;

    s = "9223372036854775807";
    number = boost::lexical_cast<double>(s);
    integer = static_cast<uint64_t>(number);
    std::cout << s << " == " << integer << std::endl;
    return 0;
}
```

lib/src/config.cc
```diff
@@ -304,6 +311,88 @@ namespace hocon {
...
+            int64_t millis = static_cast<int64_t>(number * 1000);
+            return milliseconds(millis);
+        } else if (unit_string == "h" || unit_string == "hours") {
+            int64_t millis = static_cast<int64_t>(number * 3600000);
+            return milliseconds(millis);
+        } else if (unit_string == "m" || unit_string == "minutes") {
+            int64_t millis = static_cast<int64_t>(number * 60000);
+            return milliseconds(millis);
+        } else {
+            throw bad_value_exception(*origin_for_exception, path_for_exception,
```
MikaelSmith: I don't think boost::chrono is the correct equivalent to java.time.Duration. boost::chrono is more about representing program execution time (clock cycles), not physical time. I'd look at boost::posix_time::duration instead. The boost::date_time library has been [proposed](http://www.crystalclearsoftware.com/libraries/date_time/proposal_v75/date_time_standard.html) for inclusion in the standard, but I don't think it's scheduled.

posix_time provides a 96-bit nanosecond resolution option, and facilities for translating between different units.

CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ add_definitions(${LEATHERMAN_DEFINITIONS})
 include(leatherman)
 
 # Add other dependencies
-find_package(Boost 1.54 REQUIRED COMPONENTS thread date_time chrono system program_options)
+find_package(Boost 1.54 REQUIRED COMPONENTS thread date_time chrono system program_options regex)
```
MikaelSmith: Seems like you stopped using regex, so shouldn't be needed.

lib/inc/hocon/config.hpp
```diff
@@ -594,7 +599,34 @@ namespace hocon {
...
+         */
+        virtual boost::chrono::milliseconds get_duration(std::string const& path) const;
+
+        /**
+         * Gets a value as a number of the specified units.
+         * @param path the path to the time value
+         * @param unit the units of the number returned
+         * @return the value converted to the requested units
+         */
+         virtual double get_duration(std::string const& path, time_unit unit) const;
```
MikaelSmith: It might be nice to also provide a 64-bit int version of this.

MikaelSmith:
@branan the original version of these utilities provides a 64-bit int type, and can parse both positive and negative durations. So I'd support adding a `int64_t` helper. I also think we should provide a 96-bit precision interface via Boost.Date_time's `posix_time::time_duration` class, instead of `boost::chrono::duration`.

lib/src/config.cc
branan: We actually can't rely on 80-bit precision on x86 anyway - it's pretty common for the compiler to decide to use SSE, at least on 64-bit (the standard Unix ABI actually specifies passing floating-point arguments in XMM registers, although internal to functions the optimizer will do whatever it sees fit). We could specify `long double` explicitly, but we can't really rely on having 80-bit floats on non-intel hardware anyway.

I'm also kind of OK losing precision if people write ridiculous intervals - this isn't designed to input precise dates, and the sorts of durations that we're anticipating needing can certainly be "fuzzy". If someone expects our software to respect a very large & precise number of ns, we probably can't make them happy anyway

MikaelSmith: I'd try to keep this issue in mind while checking out Boost.Date_time. If we still need to manually parse the string, I'd rather ensure it tries a 64-bit integer first, then double.

branan: Yeah. It's probably worth trying to parse as an integer first so we don't lose precision if the user is giving us a 64-bit representable number.

Really we should parse into a bigdecimal that we can truncate correctly, but unless Boost provides that for us I still don't think it's worth the implementation effort to be a little more precise on weird inputs

lib/inc/hocon/config.hpp
Magisus: How should that handle "casting up" e.g. the original value was 10 seconds, and the user requests days?

MikaelSmith: Truncation seems reasonable (i.e. 10 seconds -> 0 days).

Magisus: That looks like what the Java does (https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/TimeUnit.html#convert(long,%20java.util.concurrent.TimeUnit)). Will do.

Magisus: What is the correct way to handle this "overload"?

For the record, the Java only provides the integer version.

lib/tests/config_test.cc
```diff
@@ -116,39 +116,45 @@ TEST_CASE("correct exceptions should be thrown", "[config]") {
...
+        REQUIRE(2 == config::parse_duration("2millis", fake_origin(), "millis").total_milliseconds());
+        REQUIRE(2 == config::parse_duration("2millisecond", fake_origin(), "millis").total_milliseconds());
+        REQUIRE(2 == config::parse_duration("2s", fake_origin(), "seconds").seconds());
+        REQUIRE(2 == config::parse_duration("2seconds", fake_origin(), "seconds").seconds());
+        REQUIRE(2 == config::parse_duration("2m", fake_origin(), "minutes").minutes());
+        REQUIRE(2 == config::parse_duration("2minutes", fake_origin(), "minutes").minutes());
+        REQUIRE(2 == config::parse_duration("2h", fake_origin(), "hours").hours());
+        REQUIRE(2 == config::parse_duration("2 hours", fake_origin(), "hours").hours());
+        REQUIRE(48 == config::parse_duration("2d ", fake_origin(), "days").hours());
+        REQUIRE(48 == config::parse_duration(" 2days", fake_origin(), "days").hours());
```
MikaelSmith: Can you add some tests around bounds checking? Like 2^63 nanoseconds and some large number of days?

CMakeLists.txt
```diff
@@ -34,6 +34,9 @@ include(leatherman)
 # Add other dependencies
 find_package(Boost 1.54 REQUIRED COMPONENTS thread date_time chrono system program_options)
 
+# Add nanosecond resolution to boost::posix_time::time_duration
+add_definitions(-DBOOST_DATE_TIME_POSIX_TIME_STD_CONFIG)
```
MikaelSmith: I have concerns about this as a library. I think we need to document it in config.hpp and the README, and define in config.hpp for those consuming hocon as a library.

If it becomes a problem for someone, we can later make this a compile-time flag and have split implementation depending on the flag.

lib/src/config.cc
```diff
@@ -304,6 +310,130 @@ namespace hocon {
...
+        } else if (auto str = dynamic_pointer_cast<const config_string>(v)) {
+            return parse_duration(str->transform_to_string(), str->origin(), path);
+        } else {
+            throw bad_value_exception(*v->origin(), path, "Value at '" + path + "' was not a number or string.");
+        }
+    }
+
+    double config::get_duration_as_double(string const& path, time_unit unit) const {
+        auto timespan = get_duration(path);
+        int64_t nanos = timespan.total_nanoseconds();
```
MikaelSmith: There's really no conversion here? Converting nanoseconds to a 64-bit number is lossy with the config option above.

lib/src/config.cc
```diff
@@ -304,6 +310,130 @@ namespace hocon {
...
+            default:
+                throw config_exception("Not a valid time_unit");
+        }
+    }
+
+    int64_t config::get_duration_as_long(string const& path, time_unit unit) const {
+        auto timespan = get_duration(path);
+        int64_t nanos = timespan.total_nanoseconds();
+        switch (unit) {
+            case time_unit::NANOSECONDS:
```
MikaelSmith: Converting nanoseconds to a 64-bit number is lossy with the config option above. `time_duration` defines `total_nanoseconds`, `total_microseconds`, `total_milliseconds`, `total_seconds`. `minutes` and `hours` exist as well, but don't truncate, so I'm not sure they're the best. But it's better to do math off `total_seconds` in that case.

Magisus: So originally I was using those, but they were behaving incorrectly. Initializing the time_duration with the max nanosecond value and then calling total_seconds on it returned a completely incorrect number. And I'm not really sure why. But calling total_nanoseconds always returns the right value, so I did the math off of that one.

lib/src/config.cc
Magisus: Which config option? And what conversion were you hoping for?

lib/src/config.cc
```diff
@@ -304,6 +310,130 @@ namespace hocon {
...
+                    decimal *= 1000.0;
+                } else if (unit_string.empty() || unit_string == "ms" || unit_string == "millis" || unit_string == "milliseconds") {
+                    decimal *= 1000000.0;
+                } else if (unit_string == "s" || unit_string == "seconds") {
+                    decimal *= 1000000000.0;
+                } else if (unit_string == "m" || unit_string == "minutes") {
+                    decimal *= 60000000000.0;
+                } else if (unit_string == "h" || unit_string == "hours") {
+                    decimal *= 3600000000000.0;
+                } else if (unit_string == "d" || unit_string == "days") {
```
MikaelSmith: This has so many overflow possibilities, but Im not really sure what to do with it given the interface we're provided. If you're using decimal time then it's your problem.

lib/src/config.cc
MikaelSmith: Ugh... want to dig into the boost code? I guess I can live with this for now.

Magisus: It spits out 633437444 (0x25c17d04), when it should be about 9223372036. Does that first number look familiar?

MikaelSmith: Nope... not familiar.

lib/src/config.cc
```diff
@@ -305,6 +311,171 @@ namespace hocon {
...
+        } else {
+            throw bad_value_exception(*v->origin(), path, _("Value at '{1}' was not a number or string.", path));
+        }
+    }
+
+    double config::get_duration_as_double(string const& path, time_unit unit) const {
+        auto timespan = get_duration(path);
+        switch (unit) {
+            case time_unit::NANOSECONDS:
+                return (timespan.first * 1000000000.0) + timespan.second;
```
MikaelSmith: It would be nice to throw an exception on overflow/underflow. See http://stackoverflow.com/questions/15655070/how-to-detect-double-precision-floating-point-overflow-and-underflow for some ideas.

MikaelSmith:
👍 I like this approach best so far.

Magisus:
So are we alright with the fact that although we can parse and return decimal numbers, we can only handle durations in the range of 64-bit signed integer's worth of seconds? It's at least well-defined and easy to document.

MikaelSmith:
I'd just as well drop returning decimal numbers, so I'm also ok with that restriction.

MikaelSmith:
👍

Magisus:
Any cases in particular you want to see besides these three: https://github.com/puppetlabs/cpp-hocon/pull/71/files#diff-bafa1c1d4523a05d4e2e67f9e7c0b03dR157

lib/tests/config_test.cc
```diff
@@ -113,3 +114,49 @@ TEST_CASE("correct exceptions should be thrown", "[config]") {
...
+        REQUIRE(43200.0 == conf->get_duration_as_double("durations.secondWithFractional", time_unit::MILLISECONDS));
+        REQUIRE(-4.0 == conf->get_duration_as_double("durations.minusSeconds", time_unit::SECONDS));
+        REQUIRE(0.5 == conf->get_duration_as_double("durations.halfSecond", time_unit::SECONDS));
+        REQUIRE(1000.0 == conf->get_duration_as_double("durations.second", time_unit::MILLISECONDS));
+        // getting as a double retains fractional part
+        REQUIRE((1.0 / 60.0) == conf->get_duration_as_double("durations.minute", time_unit::HOURS));
+    }
+
+    SECTION("should throw an exception when overflow occurs", "[config]") {
+        REQUIRE_THROWS(conf->get_duration_as_long("durations.largeDays", time_unit::DAYS));
```
MikaelSmith: Why does this overflow?

MikaelSmith:
Nope, just missed them the first time.

lib/tests/config_test.cc
Magisus: Because 2^63 days is longer than 2^63 seconds, which is the max we can support.

PR #72, opened 2016-08-21
(HC-86) Generate msg catalogs, review and mark translatable strings

ahenroid:
This update marks a variety of exception and error message for
translation, converting formatted strings from boost::format style
%N% to boost::locale::format style {N}, and updates the build to
externalize strings with xgettext (generate cpp-hocon.pot).


appveyor.yml
```diff
@@ -1,11 +1,12 @@
 environment:
-  LEATHERMAN_VERSION: 0.4.2
+  LEATHERMAN_VERSION: 0.9.0
 install:
   - git submodule update --init --recursive
 
   - choco install -y mingw-w64 -Version 4.8.3 -source https://www.myget.org/F/puppetlabs
   - choco install -y cmake -Version 3.2.2 -source https://www.myget.org/F/puppetlabs
-  - SET PATH=C:\tools\mingw64\bin;%PATH%
+  - choco install -y gettext -Version 0.19.6 -source https://www.myget.org/F/puppetlabs
```
MikaelSmith: You'll also want to update .travis_target.sh to install a recent version of gettext, as in https://github.com/puppetlabs/leatherman/blob/master/scripts/travis_target.sh#L21-L25

lib/inc/hocon/config_exception.hpp
```diff
@@ -12,10 +16,10 @@ namespace hocon {
      */
     struct config_exception : public std::runtime_error {
         config_exception(config_origin const& origin, std::string const& message) :
-                runtime_error(origin.description() + ": " + message) { }
+                runtime_error(_("{1}: {2}", origin.description(), message)) { }
```
adrienthebo: Would it make more sense to translate `origin.description()` and `message`? Or are those already translated?

MikaelSmith: Origin is user-specified (usually a file name). Message appears to be translated at the calls to the config_exception constructor. See line 33 for 1 example.

ahenroid: Yes, I just did the same search, and the messages to config_exception all appear marked/translated.

Magisus:
Looks good to me 👍

PR #73, opened 2016-08-29
(maint) Fix i18n support and header file declarations

ahenroid:
This fix replaces "using leatherman::locale::_" references in cpp-hocon
header files. Placed in headers, these declarations pollute the namespace
of components, like facter, that include these headers.


Magisus:
👍

PR #77, opened 2016-09-12
(maint) Allow unregistered options in boost::program_options interface

Magisus:
Parsing more complicated HOCON structures (e.g. a list of HOCON objects) into boost::program_options does not work very well. This allows `program_options` to ignore keys not specified in the `options_description`, so that they can instead be handled specially.

The second commit adds a useful method to the public API of config_object, to make interacting with objects easier. This method already existed in the most common concrete implementation of the abstract class; this makes the method publicly accessible without casting.

### DIFF
--- a/lib/tests/program_options.cc
+++ b/lib/tests/program_options.cc
@@ -31,14 +31,30 @@ TEST_CASE("arrays with only values are converted to boost::po") {
     REQUIRE(vm.count("foo") == 1);
 }
 
-TEST_CASE("arrays with objects are converted to paths in boost::po") {
+TEST_CASE("unregistered keys cause an exception when `allow_unregistered` is false") {
+    po::options_description opts("");
+    opts.add_options()("foo", "description");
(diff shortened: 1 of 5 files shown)


lib/tests/program_options.cc
```diff
@@ -31,14 +31,30 @@ TEST_CASE("arrays with only values are converted to boost::po") {
...
+
+TEST_CASE("arrays with objects cause an exception in boost::po") {
     po::options_description opts("");
     opts.add_options()("foo", "description");
     po::variables_map vm;
     REQUIRE_THROWS(po::store(parse_string(string("{foo : [ { bar : baz } ]}"), opts), vm));
 }
 
-TEST_CASE("Arrays mixed with objects produce both a field and a nested path in boost:po") {
+TEST_CASE("Arrays mixed with objects cause an exception in boost:po") {
```
branan: good catch on the misnamed test

PR #78, opened 2016-09-21
(maint) Fix linking errors when building with Leatherman locale

Magisus:
When using leatherman::locale, it is necessary to also link against
boost::locale. On OSX, this further requires linking against the iconv
library.

@MikaelSmith would you mind looking over this and seeing if any of the added bits jump out at you as being definitely unnecessary. This went through enough iterations that I think some things may have been added that are actually irrelevant (and if you find any, I'd love to hear why they're not needed. Still trying to get the full picture on this stuff).

### DIFF
--- a/lib/tests/CMakeLists.txt
+++ b/lib/tests/CMakeLists.txt
@@ -30,6 +30,10 @@ target_link_libraries(lib${PROJECT_NAME}_test
     ${CMAKE_THREAD_LIBS_INIT}
 )
 
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND BOOST_STATIC)
+    target_link_libraries(lib${PROJECT_NAME}_test iconv)
+endif()
+
 add_test(NAME "unit_tests" COMMAND lib${PROJECT_NAME}_test)
 
 configure_file (
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ add_definitions(-DLEATHERMAN_I18N)
 include(leatherman)
 
 # Add other dependencies
-find_package(Boost 1.54 REQUIRED COMPONENTS thread date_time chrono system program_options)
+find_package(Boost 1.54 REQUIRED COMPONENTS locale thread date_time chrono system program_options)
 
 # pthreads if you require the Boost.Thread component.
 find_package(Threads)


MikaelSmith:
Yeah, I think this is unfortunately required given how the leatherman/locale/locale.hpp header is structured.

lib/tests/CMakeLists.txt
```diff
@@ -18,16 +18,22 @@ set(TEST_CASES
...
 
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin" AND BOOST_STATIC)
+    target_link_libraries(lib${PROJECT_NAME}_test iconv)
+endif()
+
 target_link_libraries(lib${PROJECT_NAME}_test
     ${Boost_LIBRARIES}
-    ${LEATHERMAN_LIBRARIES}
     ${CMAKE_THREAD_LIBS_INIT}
+    ${LEATHERMAN_LIBRARIES}
```
MikaelSmith: Why was this change necessary?

MikaelSmith: Link order matters. I'd put iconv after Boost_LIBRARIES (probably doesn't matter where it is related to LEATHERMAN_LIBRARIES, but after them as well makes sense to me).

lib/tests/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # Setup compiling the test executable. C++ compile flags are inherited from the parent directory.
-include_directories(${LEATHERMAN_CATCH_INCLUDE})
+include_directories(../inc ${LEATHERMAN_CATCH_INCLUDE} ${Boost_INCLUDE_DIRS})
```
MikaelSmith: How did this work before?

lib/tests/CMakeLists.txt
```diff
@@ -18,16 +18,22 @@ set(TEST_CASES
     program_options.cc
 )
 
+link_directories(${Boost_LIBRARY_DIRS})
```
MikaelSmith: This should be implied by find_package(Boost).

lib/CMakeLists.txt
```diff
@@ -70,6 +70,7 @@ set(PROJECT_SOURCES
     src/values/config_concatenation.cc
     src/config_value_factory.cc)
 
+link_directories(${Boost_LIBRARY_DIRS})
```
MikaelSmith: Should also be unnecessary.

lib/tests/CMakeLists.txt
MikaelSmith: Oh, the changes here aren't needed, they're inherited from the parent (lib/CMakeLists.txt).

Magisus:
Looks like it still works with the changes you suggested.

PR #79, opened 2016-09-22
(maint) Fix linking errors on AIX, Solaris, Cisco, and Windows

Magisus:
This has been run against a full platform set, with successful builds on all of them. Before this can be included in puppet-agent, Facter will need to be updated with similar conditionals around boost::locale.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,12 +28,24 @@ include(options)
 include(cflags)
 set(${PROJECT_NAME_UPPER}_CXX_FLAGS "${LEATHERMAN_CXX_FLAGS}")
 add_definitions(${LEATHERMAN_DEFINITIONS})
-add_definitions(-DLEATHERMAN_I18N)
+
+if(LEATHERMAN_USE_LOCALES)
+    add_definitions(-DLEATHERMAN_I18N)
+endif()
+
 ## Pull in helper macros for working with leatherman libraries
 include(leatherman)
 
+if(LEATHERMAN_USE_LOCALES)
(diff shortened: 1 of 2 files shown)


lib/CMakeLists.txt
```diff
@@ -80,7 +80,6 @@ add_library(lib${PROJECT_NAME} $<TARGET_OBJECTS:libprojectsrc>)
 set_target_properties(lib${PROJECT_NAME} PROPERTIES VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
 target_link_libraries(lib${PROJECT_NAME}
     ${LEATHERMAN_LIBRARIES}
-    ${Boost_LIBRARIES}
```
Magisus: It surprised me too. This one fixed one of the multiple definitions errors on Windows when linking Facter against cpp-hocon...

CMakeLists.txt
```diff
@@ -28,12 +28,24 @@ include(options)
...
+    add_definitions(-DLEATHERMAN_I18N)
+endif()
+
 ## Pull in helper macros for working with leatherman libraries
 include(leatherman)
 
+SET(BOOST_COMPONENTS thread date_time chrono system program_options)
+
+if(LEATHERMAN_USE_LOCALES)
+    list(APPEND BOOST_COMPONENTS locale)
```
Magisus: Updated

Magisus: updated

PR #83, opened 2016-10-19
(HC-95) Fix CPP builds on FreeBSD 10.3 with system clang

jmslagle:


MikaelSmith:
@magisus any chance you could check that this enables us to remove the clang patches?

branan:
@MikaelSmith @Magisus it ought to enable that, yeah. I was working with Jason on this at puppetconf last week.

PR #86, opened 2016-10-25
(HC-97) Remove library prefixing.  Add support to build libcpp-hocon …

jmslagle:
…shared

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -75,20 +75,23 @@ set(PROJECT_SOURCES
 ## Without the intermediate target, unexported symbols can't be tested.
 add_library(libprojectsrc OBJECT ${PROJECT_SOURCES})
 set_target_properties(libprojectsrc PROPERTIES POSITION_INDEPENDENT_CODE true)
-
-add_library(lib${PROJECT_NAME} $<TARGET_OBJECTS:libprojectsrc>)
-set_target_properties(lib${PROJECT_NAME} PROPERTIES VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
-target_link_libraries(lib${PROJECT_NAME}
+if(CPP_HOCON_SHARED)
+	add_library(${PROJECT_NAME} SHARED $<TARGET_OBJECTS:libprojectsrc>)
+else()
+	add_library(${PROJECT_NAME} $<TARGET_OBJECTS:libprojectsrc>)
+endif()
+set_target_properties(${PROJECT_NAME} PROPERTIES VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
+target_link_libraries(${PROJECT_NAME}
     ${LEATHERMAN_LIBRARIES}
     ${Boost_LIBRARIES}
     )
 
 # Generate the export header for restricting symbols exported from the library.
 # Restricting symbols has several advantages, noted at https://gcc.gnu.org/wiki/Visibility.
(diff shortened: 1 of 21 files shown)


lib/CMakeLists.txt
```diff
@@ -75,20 +75,23 @@ set(PROJECT_SOURCES
 ## Without the intermediate target, unexported symbols can't be tested.
 add_library(libprojectsrc OBJECT ${PROJECT_SOURCES})
 set_target_properties(libprojectsrc PROPERTIES POSITION_INDEPENDENT_CODE true)
-
-add_library(lib${PROJECT_NAME} $<TARGET_OBJECTS:libprojectsrc>)
-set_target_properties(lib${PROJECT_NAME} PROPERTIES VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
-target_link_libraries(lib${PROJECT_NAME}
+if(CPP_HOCON_SHARED)
```
MikaelSmith: This is actually unnecessary, you can use https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html

lib/src/config_document_parser.cc
```diff
@@ -33,8 +33,7 @@ namespace hocon { namespace config_document_parser {
 
     /** Parse context */
     parse_context::parse_context(config_syntax flavor, shared_origin origin, token_iterator tokens) :
-        _line_number(1), _tokens(move(tokens)), _flavor(flavor), _base_origin(move(origin)),
-        _equals_count(0) { }
```
MikaelSmith: Why is this removed? It should still be initialized (C++ doesn't default initialize ints to 0).

MikaelSmith:
Thanks for working on this fix!

lib/CMakeLists.txt
jmslagle: Aha!  I can nuke that - I just copied what Leatherman is going :)

lib/src/config_document_parser.cc
jmslagle: Erm?  I actually have no clue where those went.  I'm guessing they accidently got nuked as I went through and fixed all the LIBCPP atuff.  I'll fix.

jmslagle:
I've made the changes.  I suspect it's just an oddity with when I pulled my initial branch for HC-95 on the variable thing.  The CMake trickery I didn't know and am updating my port.

paravoid:
I basically did the very same, unaware of this :) This is the fallout from puppetlabs/leatherman#213 which in turn is pretty much required for packaging them for Debian.

MikaelSmith:
@puppetlabs/night-s-watch do you have some bandwidth to deal with any downstream effects of this? I think everything will still find the library, but not 100% sure.

GabrielNagy:
Hi @MikaelSmith,

I started an agent build with this change and https://github.com/puppetlabs/leatherman/pull/213, everything compiles fine (facter, leatherman, libwhereami) until cpp-pcp-client which gets installed to `liblibcpp-pcp-client.so` and pxp-agent cannot find it.

I assume something similar must be also done on cpp-pcp-client?

GabrielNagy:
@smortex unfortunately this change greatly impacts how we build our puppet-agent packages.

Since we're looking to remove cpp projects from our pipeline (as we're introducing [facter-ng](https://github.com/puppetlabs/facter-ng/)) I'm not sure we want to make these changes (impacted projects are cpp-hocon, cpp-pcp-client, leatherman, libwhereami, facter, and still I ran into some issues with libfacter.so not finding libraries).

GabrielNagy:
@jmslagle thanks for the understanding, I'll go ahead and close this then.

PR #89, opened 2016-11-11
(doc) Fix README requirements

MikaelSmith:
The README requirements for GCC, CMake, and Boost were inconsistent with
actually use. Update to reflect that we build with GCC 4.8, and the
CMake code requires CMake 3.2.2 and Boost 1.54.


MikaelSmith:
@Magisus mind checking and merging?

README.md
```diff
@@ -41,9 +41,9 @@ M=MMMMMMMM++  FOR HUMANS   ++M8MMMMMM7M
 ## Build Requirements
 
 * OSX or Linux
-* GCC >= 5.0 or Clang >= 3.4 (with libc++)
-* CMake >= 3.0
-* Boost Libraries >= 1.57
+* GCC >= 4.8 or Clang >= 3.4 (with libc++)
+* CMake >= 3.2.2
+* Boost Libraries >= 1.54
```
Magisus: The lowest Boost version I'm seeing is 1.55 (in Travis). Do we still build with 1.54 somewhere? pl-build-tools looks like 1.58, but I could be missing something.

MikaelSmith: I don't think we do, but as far as I'm aware 1.54 is still sufficient. That's what the Facter docs currently say. We require 1.54 for Boost.Log.

Magisus: Alright, sounds good.

Magisus:
👍  will merge when check complete

MikaelSmith:
I used skip ci, so it won't run.

Magisus:
Oh weird, it's looking like it's still trying *shrug* Apparently I have admin rights for this one though, so... merging!

PR #93, opened 2017-09-15
(maint) Rebuild AppVeyor artifacts, release as 0.1.6

MikaelSmith:
Rebuilds AppVeyor CI release artifacts with GCC 5.2.0.


MikaelSmith:
Should be able to do the same for Facter once this is done.

PR #97, opened 2017-09-29
Resolve included files before including them

### LINKED ISSUE #95: included file root
Continuation of the #94, it's different problem but the setup is the same (so skipping the Scala/C++ code).

if we have the config files:
test-1.conf:
```
Foo {
    foo = "foo"
}

Bar {
    include file("bar")                                                                                                                                                                        
}
```
bar.conf:
```
bar = "bar"
doubleBar = ${bar}
```
it fails the substitution with: "Could not resolve subtitution to a value: ${bar}".

If we use instead:
```
doubleBar = ${Bar.bar}
```
it works and the substitution is performed.

In the reference implementation (typesafe-config) both cases succeed.

luben:
Fixes #95 : substitution inside the file should work without the full
path where they are included.

### DIFF
--- a/lib/src/simple_includer.cc
+++ b/lib/src/simple_includer.cc
@@ -58,7 +58,7 @@ namespace hocon {
     }
 
     shared_object simple_includer::include_file_without_fallback(shared_include_context context, std::string what) {
-        return config::parse_file_any_syntax(move(what), context->parse_options())->root();
+        return config::parse_file_any_syntax(move(what), context->parse_options())->resolve(config_resolve_options(true, true))->root();
     }
 
     config_parse_options simple_includer::clear_for_include(config_parse_options const& options) {


MikaelSmith:
Ah, sorry I missed this.

Can you add a test like https://github.com/puppetlabs/cpp-hocon/blob/master/lib/tests/conf_parser_test.cc#L627-L633 - but with `file()` removed - to the PR?

MikaelSmith:
Also, thanks so much for investigating and putting together a fix!

luben:
This PR does not fix `include file("path")` vs `include "path"` case but it's about resolving the substitutions in the included files: don't only consider path from the config root but also from the included file root.

MikaelSmith:
Ah right, that would require a more complex example.

luben:
Actually the fixtures already had an example pattern I can use. Sent a PR with the test,

Issue #105, opened 2018-04-24
Self-Referential Substitutions don't work


troizky:
I'm using the example from documentation: https://github.com/lightbend/config/blob/master/HOCON.md#examples-of-self-referential-substitutions
```
A self-reference resolves to the value "below" even if it's part of a path expression. So for example:

foo : { a : { c : 1 } }
foo : ${foo.a}
foo : { a : 2 }

Here, ${foo.a} would refer to { c : 1 } rather than 2 and so the final merge would be { a : 2, c : 1 }.
Recall that for a field to be self-referential, it must have a substitution or value concatenation as its value. If a field has an object or array value, for example, then it is not self-referential even if there is a reference to the field itself inside that object or array.
```
And I'm getting crash:
`
terminate called after throwing an instance of 'hocon::bug_or_broken_exception'
  what():  resolve_substitutions() did not give us a resolved object
`
Also I'm getting crash in cases below:
```
a = {aa = abba}
a.aa = mama
b = ${a} { bb = mia }
# Everything below is failing
b.aa = mea
b.bb = culpa
b = { aa = mea, bb = culpa }
c = ${b.aa}
c = ${b.bb}
```
Looks like this library implementation doesn't follow the HOCON documentation...

MikaelSmith:
The implementation is incomplete. I would appreciate any help finishing it for those cases.

troizky:
Even the simplest testcase does not work:
```
  "a": {
    "b": 27
  },
  "obj": ${a}
  "val": ${obj.b}
```
Maybe this one is easier to fix?

MikaelSmith:
That is likely [HC-73](https://tickets.puppetlabs.com/browse/HC-73). I'll take a quick look to see if I can figure out what's wrong, but don't have a ton of time to spend on it before I'm on vacation for 2 weeks.

MikaelSmith:
This library was implemented to closely mirror the structure of https://github.com/lightbend/config. So tracing the exception I get with your simple test case back to https://github.com/puppetlabs/cpp-hocon/blob/master/lib/src/values/config_reference.cc#L75 and comparing to https://github.com/lightbend/config/blob/master/config/src/main/java/com/typesafe/config/impl/ConfigReference.java#L70, it looks like the fallback code at https://github.com/puppetlabs/cpp-hocon/blob/master/lib/src/values/config_reference.cc#L61 wasn't implemented. I'll try that out.

MikaelSmith:
[...]

PR #106, opened 2018-05-17
Fixed Self-Reference Substitution

troizky:
According to documentation: https://github.com/lightbend/config/blob/master/HOCON.md#examples-of-self-referential-substitutions
```
A self-reference resolves to the value "below" even if it's part of a path expression. So for example:

foo : { a : { c : 1 } }
foo : ${foo.a}
foo : { a : 2 }

Here, ${foo.a} would refer to { c : 1 } rather than 2 and so the final merge would be { a : 2, c : 1 }.
Recall that for a field to be self-referential, it must have a substitution or value concatenation as its value. If a field has an object or array value, for example, then it is not self-referential even if there is a reference to the field itself inside that object or array.
```
But I'm getting crash in cases below:
```
"a": {"b": 27},
"obj": ${a}
# below is failing
"val": ${obj.b}

a = {aa = mama}
b = ${a} { bb = mia }
# below is failing
b.aa = mea
b.bb = culpa
```
Open Issue https://github.com/puppetlabs/cpp-hocon/issues/105

### DIFF
--- a/lib/tests/config_substitution_test.cc
+++ b/lib/tests/config_substitution_test.cc
@@ -112,6 +112,34 @@ static shared_object subst_env_var_object() {
     return resolved;
 }
 
+static shared_object subst_nested_chain_object() {
+    static auto const resolved = parse_object(R"(
+{
+  "a": {
+    "b": 729
+  },
+  a.b: 27,
+  "obj": ${a}
+  "val": ${obj.b}
+}
+)");
+
+    return resolved;
+}
+
+static shared_object subst_nested_assign_object() {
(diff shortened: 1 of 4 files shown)


lib/tests/config_substitution_test.cc
```diff
@@ -112,34 +112,6 @@ static shared_object subst_env_var_object() {
     return resolved;
 }
 
-static shared_object subst_nested_chain_object() {
```
MikaelSmith: Why remove these tests? Decided they duplicated others?

MikaelSmith:
Thanks so much for looking into this!

PR #107, opened 2018-05-18
Added config resolver

troizky:
As it was suggested here https://github.com/puppetlabs/cpp-hocon/issues/105
Comparing to https://github.com/lightbend/config/blob/master/config/src/main/java/com/typesafe/config/impl/ConfigReference.java#L93, it looks like the fallback code at https://github.com/puppetlabs/cpp-hocon/blob/master/lib/src/values/config_reference.cc#L61 wasn't implemented.

### DIFF
--- a/lib/inc/hocon/config_resolve_options.hpp
+++ b/lib/inc/hocon/config_resolve_options.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "types.hpp"
 #include "export.h"
 
 namespace hocon {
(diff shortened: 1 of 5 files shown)


MikaelSmith:
Initial thought, is there a test case that covers these changes?

MikaelSmith:
With the fallback code unused, and no tests, I don't feel comfortable merging this. If we find we need it in the future let's re-open it.

PR #114, opened 2019-01-21
Included file path is based on including file directory

SeaTalk:
This change is motivated by incorrect outputs when I use cpp-hocon to parse our conf files. 
[U can find more details by clicking this link about what situation I ran into.](https://github.com/SeaTalk/cpp-hocon/blob/dev/README.md) 
I wonder whether someone ever been where I am, and I think may be this change could also be a solution to them.


SeaTalk:
No report from trvis-ci yet. Don't know what I am supposed to do. 🤔

MikaelSmith:
I'll check up on Travis CI.

MikaelSmith:
Our linter is complaining about a few things.

README.md
```diff
@@ -9,6 +9,31 @@ This is a port of the TypesafeConfig library to C++.
 
 The library provides C++ support for the [HOCON configuration file format](https://github.com/typesafehub/config/blob/master/HOCON.md).
 
+## I Made Some Change
```
MikaelSmith: This probably makes more sense as the commit message rather than part of the README.

example/bin/conf_test.cpp
```diff
@@ -0,0 +1,26 @@
+#include <hocon/parser/config_document_factory.hpp>
```
MikaelSmith: This should be moved to the `lib/test` directory and integrated with our test suite.

example/conf/a.conf
```diff
@@ -0,0 +1,10 @@
+{
```
MikaelSmith: All of `example/conf` and `imgs/` seem like they should be part of the test suite as well.

lib/src/config.cc
```diff
@@ -29,13 +30,27 @@ using namespace std;
 
 namespace hocon {
 
-    shared_config config::parse_file_any_syntax(std::string file_basename, config_parse_options options) {
+    shared_config config::parse_file_any_syntax(std::string file_basename, config_parse_options options, shared_full_current fpath) {
+        std::string file_dir(""), file_name("");
+        if (nullptr == fpath) {
+            fpath = make_shared<full_path_operator>();
+        }
+        extract_filename_from_path(file_basename, &file_dir, &file_name);
```
MikaelSmith: My instinct is that this isn't the place we should be fixing this, but it'd take some work for me to follow up on.

From a quick glance at https://github.com/lightbend/config/blob/7b9e1539dd46385e9b83be43c4112a448129ac2f/config/src/main/java/com/typesafe/config/impl/Parseable.java#L639-L655, I think the mistake may be in not overriding `relative_to` on `parseable_file` (and possibly others) in https://github.com/puppetlabs/cpp-hocon/blob/master/lib/src/parseable.cc.

SeaTalk: > I left some pretty high-level comments. Do you have an interest in trying to improve this, or should we consider this an open problem for someone to build on your solution?

Thanks for code-reviewing and comments. I think we could left the problem open before solved. And I'll keep trying to figure another way out to fix it. Thanks again~ 
It's 3 in the morning and I need some sleep. Good day!

MikaelSmith:
Closed in favor of the other PR.

PR #115, opened 2019-01-29
relative file path of a include statement relatives to directory of including file

SeaTalk:
This change is another solution for  [#114](#114 )


SeaTalk:
Anyone can review this change ?

MikaelSmith:
This looks like it could work. Did you explore my idea in https://github.com/puppetlabs/cpp-hocon/pull/114#discussion_r249910325 and find it didn't help?

MikaelSmith:
This generally seems good. @caseywilliams interested in doing a review?

SeaTalk:
> This looks like it could work. Did you explore my idea in [#114 (comment)](https://github.com/puppetlabs/cpp-hocon/pull/114#discussion_r249910325) and find it didn't help?

I followed your idea to find out the key issue exists the isolation of context for each parseable instance. The parseable_file of included file has not any information from including file. I didn't override function "relative_to" for sub-classes of parseable since we don't have a particular  path for "resource", like java, in c++ for now.

lib/tests/simple_confs/a.conf
```diff
@@ -0,0 +1,11 @@
+{
+    include file("d.conf")
+    Peter : {
+        include file("b.conf")
+        include file("sub/b.conf")
+        # passwd4 : ${new_passwd}
```
MikaelSmith: This didn't work?

lib/src/parseable.cc
```diff
@@ -105,17 +105,38 @@ namespace hocon {
         // if it starts with "/" then remove the "/", for consistency
```
MikaelSmith: Looks like this comment is no longer accurate.

lib/tests/simple_confs/a.conf
SeaTalk: This works。I don't remember what I was testing that I put a sharp in front of it.

lib/src/parseable.cc
SeaTalk: yeah, I'll try to correct it

lib/inc/internal/simple_includer.hpp
```diff
@@ -56,8 +56,31 @@ namespace hocon {
...
                                                    config_parse_options parse_options) const = 0;
+
+        void set_context(shared_include_context context) {
+            if (!_init_finished) {
+                _init_finished = true;
+                _context = context;
+            }
+        }
+
+        bool context_initialed() {
```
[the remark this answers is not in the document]
SeaTalk: > Except for one spelling thing, I think this looks okay. The test cases look thorough, so 👍

Thanks for reviewing. I'll correct that ASAP.

SeaTalk:
Done. @MikaelSmith @Magisus

MikaelSmith:
Thanks so much for your work on this!

zzhang97:
I was caught by this bug too with latest release. Happily found the fix was already in master. Thanks!

PR #116, opened 2019-03-07
Fix Werror=catch-value for gcc8

greno4ka:


greno4ka:
I could not verify the email address through your form, so my contribution is trivial, just regard it and merge or deny, no matter. I just published some helpful changes, that can be interesting to other people.

PR #136, opened 2022-09-14
(PA-4632) Move travis ci to github actions

tvpartytonight:


tvpartytonight:
I have run the release branch a few times in my own fork and there appears to be race condition that I ran into once where multiple releases are created from the same tag; I have only seen it once out of 5 tags, which makes me somewhat ok with how often we might run into this issue. The workaround is pretty simply; delete the tag and push up an identical tag for the release workflow to run.

joshcooper (approved):

.github/workflows/ci_checks.yaml
```diff
@@ -0,0 +1,28 @@
+name: CI checks
+on:
+  pull_request:
+    branches: [master]
```
joshcooper, line 4: ah I noticed this check didn't run after merging and I think we want to run on push too, like we do for [leatherman](https://github.com/puppetlabs/leatherman/pull/335/files#diff-4af11422a4987e947e5a47adead7a30d32cdb2db82e2d3fe36f8e6cbe84d5ac5R5)

```yaml
  push:
    branches: [master]
```

tvpartytonight: Oh, why do we want to run it after push? I separated out the actions for push+tag into a different `release` workflow, but I didn't see any reason why these should run after merge without tags.

joshcooper: It's not strictly necessary for cpp-hocon, but in general there are a few cases where it's useful. Some of our automation will push directly, bypassing the PR+merge step. It's more of an issue for [puppet](https://github.com/puppetlabs/puppet/blob/04cf2ae3ac98d7ac628663f4ef0b652b6b013041/.github/workflows/rspec_tests.yaml#L5-L6) and [facter](https://github.com/puppetlabs/facter/blob/67f7290c2da26bc58fe2cd9d3d9c51c8b586702a/.github/workflows/unit_tests.yaml#L5-L7) because of the man page and i18n component jobs. I think there's automation to version bump cpp-hocon during releases.

Also if a PR is not based off of the latest commit in the base branch, then it's possible for a semantic conflict to occur, as in the base branch and PR pass CI, but after the merge it won't and we don't find out about it until the someone opens a new PR. It's unlikely to happen for this repo, so more of a general thing. It's also possible to guard against this in "Require branches to be up to date before merging" in the branch protection rule, but there are some downsides with that as well.

